PR #22, opened 2019-03-29
Remove line from utility.h: #include <core/dvector.h>

ant1fact:
Remove #include <core/dvector.h> to avoid compilation error with Godot's latest 3.1 stable release

After the change, the engine compiled successfully and I was able to open and run the Zylann/voxelgame project

### DIFF
--- a/utility.h
+++ b/utility.h
@@ -2,7 +2,6 @@
 #define HEADER_VOXEL_UTILITY_H
 
 #include <core/vector.h>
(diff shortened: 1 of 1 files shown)


Zylann:
Fixed in 6fc70a66c1c1de0dc8224f3d058f093292ba2d56
Also in the `dmc` branch I realized this might have been needed, so I added it back with the correct name and will be merged to master after some time.

PR #28, opened 2019-05-29
Created documentation

### LINKED ISSUE #10: Tutorial
Hey this isn't really an issue but I'm fairly new and trying to make a game using voxels, how would I go about using this module? I cloned it to the module folder but how do you set up godot from the source code? I tried making an executable using SCon but no luck. Any help is appreciated :)

tinmanjuggernaut:
This is complete as is. Someone can read this and get the source built and terrain loaded without much effort.

fixes #10


tinmanjuggernaut:
Since this is still open, please keep it open for now. I'll add the API overview to it.
Then over the next few days I'll add a page on materials and one on the current state of collision (including what I'm doing with raycasts). Then that will be enough to get anyone started.

Zylann:
The MD tutorial was OK but for the class reference I would prefer to use the same system as Godot, because Doxygen won't work well with the script API, and XML will allow to parse it into Godot Editor docs as well as MD files.

tinmanjuggernaut:
Today's force push is a complete package ready for review and perhaps merging.

The changes start at the Readme and provide navigation links throughout all the files.
https://github.com/tinmanjuggernaut/godot_voxel/blob/tutorial/README.md


Fixes #10

Zylann:
Your PR contains a `bak` folder, maybe you should remove it? (I thought Doxygen wasn't going in for now)

Also, how can I generate the docs on Windows (or any OS)? It would seem I need a Unix bash with `xsltproc` and a specific global path to the Godot executable. Can't we use Python for these things? (as it's granted as cross-platform dependency for the build system, preferably without extra deps).

Are you using this feature? https://github.com/godotengine/godot/pull/7310

In `06_custom-streams.md`, I think you should also explain that custom streams must be thread safe if they get to be used for terrain.

tinmanjuggernaut:
> Your PR contains a `bak` folder, maybe you should remove it? (I thought Doxygen wasn't going in for now)

The "for now" is the question. The bak contains the work I setup for Doxygen so it doesn't have to be redone in the future. I've removed it.



> Also, how can I generate the docs on Windows (or any OS)? It would seem I need a Unix bash with `xsltproc` and a specific global path to the Godot executable. Can't we use Python for these things? (as it's granted as cross-platform dependency for the build system, preferably without extra deps).

I've never written anything in python. If I did, we'd still have a dependency on lxml.

xlstproc is cross platform, as is bash.

**Bash**
Did you install gitforwindows.org? Then you already have a bash shell on your system. Just run the git bash icon in your start menu. Once setup, you can build the docs by running git bash, then these commands, adjusting for your path:
```
cd /c/dev/godot/modules/voxel/doc/tools
./build_api gd
```
This will create and overwrite the files in ../api. It won't delete them, so if a class has been removed, the API file will still be there.

It does require editing the variables at the top of the script to specify the path to godot.

You can also build Godot with GCC in this environment given the instructions I provided in #10 .
Other basic commands are:
* `ls` for a directory listing
* `cd` change directory
* `rm file` to delete file
* `mv file1 file2` to move or rename file1 to file2
* `echo $PATH` to print your path

**xsltproc:**
1. Go here: https://www.zlatkovic.com/pub/libxml/
1. Download `libxslt`, `libxml2`, `iconv`, `zlib`
1. Look in the bin directories for each zip. You need xsltproc.exe, xmllint.exe, and all the DLLs. Or just grab everything in all bin folders.
1. Put them in your path like c:\windows. Git bash put this in my path: c:\users\cory\bin. So I created that folder and dumped those files in there. Then from git bash, I ran `xsltproc -version` and got a version string.

[ Reference: http://www.sagehill.net/docbookxsl/InstallingAProcessor.html ]



Is this an acceptable process?



> Are you using this feature? [godotengine/godot#7310](https://github.com/godotengine/godot/pull/7310)

No, I just have godot dump all classes it's compiled with, pull out Voxel* and delete the rest. There is no module specific configuration setup for documentation (i.e. doc_classes). I was thinking it could be once the API is closer to feature complete and we want to start editing XML files with documentation. That PR mentions classes.xml. What is this? I don't see any example of that file in all of the godot source.



> In `06_custom-streams.md`, I think you should also explain that custom streams must be thread safe if they get to be used for terrain.

Noted at the top.

Zylann:
> I've never written anything in python. If I did, we'd still have a dependency on lxml.

I don't know bash scripts and my last use of xslt goes back to 10 years ago^^ Also Godot doesn't depend on lxml either to generate its docs ;) Also Python can load xml just fine built-in.
I do have Git bash for Windows but I had a few issues with it in the past to execute proper bash files, due to a few clashes with Windows I believe. I don't mind that much xslt and I might give it a try, but I really like when a repo can work out of the box with minimal dependencies like Godot does, especially when you have python around (also it seems those dependencies are not always present in Git for Windows?).

> No, I just have godot dump all classes it's compiled with, pull out Voxel* and delete the rest. There is no module specific configuration setup for documentation (i.e. doc_classes). I was thinking it could be once the API is closer to feature complete and we want to start editing XML files with documentation. That PR mentions classes.xml. What is this? I don't see any example of that file in all of the godot source.

Fair enough. `classes.xml` is the source file for Godot API documentation, which gets compiled both for Sphinx and for the editor. Tbh I didn't have a look at this process for modules yet but the idea is to be able to document modules the same way as we do in Godot (`classes.xml` has been global for a long time), so one of the goals of that PR was to be able to have that locally per module as well. If something is still missing from there we should work to improve it because that could also benefit modules that are included in Godot's repo.

tinmanjuggernaut:
I was not able to get lxml to process the xml and xslt files. :( So I wrote a custom xml formatter and a separate build script, both in python.

xmlproc.py can be run on one file:
`python xmlproc.py /c/dev/gd/godot/doc/classes/Spatial.xml`
That will print to stdout. Or a second file can be specified that it will output to.

build.py requires python 3.4+ and xmlproc.py. It can be edited to have the godot path and output dir, or it can be specified on the command line with -g. The output directory should be the default of '../api' however you can specify it with -o. -v turns on verbose mode.
``python build.py -g /c/dev/gd/Godot-3.2-629bc10d8.exe -o api-temp``

These work on windows and linux with no dependencies except python 3.4+.

doc/tools/build.py
```diff
@@ -0,0 +1,152 @@
...
+if(sys.version_info<(3,4)):
+    print("Please upgrade python to version 3.4 or higher.")
+    print("Your version: %s\n" %sys.version_info)
+    sys.exit(1)
+
+from pathlib import Path
+
+# Configure these variables for your system or specify on the command line
+
+GD=str(Path.home()) + "/bin/godot-dev"    # Path to Godot executable
```
Zylann: By default this should better be `../../bin/godot.<platform>.<opt?>.<tools?>.<bits>.exe`, which is where Godot is built. What do you think?

doc/tools/build.py
```diff
@@ -0,0 +1,152 @@
...
+
+        for i in files:
+            f = Path(i).name
+            str += "* [%s](%s)\n" % (f, f)
+
+        str += "\n---\n"
+        str += "* [Doc Index](../01_get-started.md)\n"
+        str += "_Generated on " + strftime("%b %d, %Y", gmtime()) + "_\n" 
+
+        if verbose: print("Writing %s" % index)
```
Zylann: Indent?

doc/tools/build.py
```diff
@@ -0,0 +1,152 @@
...
+from xmlproc import process_xml
+
+
+
+def build_gdscript_api(verbose=False):
+
+
+        # Make temp dirs
+        tmp_d = tempfile.mkdtemp()
+        if verbose: print ("Making temp directories in " + tmp_d)
```
Zylann: Indent?

doc/tools/build.py
```diff
@@ -0,0 +1,152 @@
...
+        str += "_Generated on " + strftime("%b %d, %Y", gmtime()) + "_\n" 
+
+        if verbose: print("Writing %s" % index)
+        outfile = open(index, 'a')
+        outfile.write(str)
+        count += 1
+
+
+	# Remove temporary files
+        shutil.rmtree (tmp_d, ignore_errors=True)
```
Zylann: I think xml files are actually useful for the engine to grab when it builds, so that the doc gets embedded in the editor as well

doc/tools/build.py
tinmanjuggernaut: Good idea for the path. If the user wants to call Godot, they don't need the first script. They can just use the xmlproc.py script themselves. It's designed for command line or import.

doc/tools/build.py
tinmanjuggernaut: Indent what? 
_Generated_?

Or indent into the for loop above? No.

Do you not like "if verbose: print" all on one line?

doc/tools/build.py
tinmanjuggernaut: if verbose: print? 
Works fine.

doc/tools/build.py
tinmanjuggernaut: This already happens. Godot already builds Voxel* classes into the online searchable help. It does so from the code (probably the GDCLASS macros or register functions). It does not need the XML files to generate online help class information.

However, if you want to provide descriptions, then you must use the --doctool to extract the XML files, hand edit them, then feed them back into the system either by putting them in the global doc/classes folder or by setting up doc_classes in the module's directory.

tinmanjuggernaut:
PR updated with new script that searches for godot in the bin directory. The only question is what you meant by indent.

doc/tools/build.py
Zylann: It's the inline block yes. I prefer it in two lines, you actually wrote two lines sometimes.

doc/tools/build.py
Zylann: That's precisely because of function descriptions. You'll notice the Godot repo versions its XML files too for that reason, so the same must be done for modules. The doctool helps update it, but it should not overwrite. That's why I opened https://github.com/godotengine/godot/issues/4562 originally :)

tinmanjuggernaut:
I have updated the indent style.

tinmanjuggernaut:
I have updated the docs to align with the collision PR. Is there anything else you are expecting before merging this PR?

Zylann:
I think I will merge it even though I need to do a more thorough pass on how Godot modules should be documented in general.
For collision, I might also merge it, but I might also change it in the future to not instanciate nodes. Apart from overhead, another reason is when you get collision results, I want it to return the terrain, not a dynamically-instanced child staticbody.

tinmanjuggernaut:
No problem. I fully suspect things will change in the future. But the docs and collision now are a really solid first step that makes the project much more accessible and useful for people. For myself it makes it usable so I can start prototyping some game ideas I have.

Zylann:
I finally had a look at how the Godot `doctool` works for modules. https://github.com/godotengine/godot/pull/7310 was supposed to support a workflow where modules could have their own doc, but it broke over time. In addition to that, module docs continue to be aggregated in global `doc/classes/` which makes the issue still stand.

I'm considering changing again doctool so that module docs actually get located in module folders, so modules like this one would be able to have their own documentation generated from `doctool` visible from the editor, while preserving handwritten descriptions.

Zylann:
So it turns out some code from that old PR was never cleaned up, and I found out the workflow changed in the docs: https://docs.godotengine.org/en/3.1/development/cpp/custom_modules_in_cpp.html#writing-custom-documentation

It's a bit inconvenient, but I'm going to set this up.

PR #29, opened 2019-06-02
Raycast for VoxelLodTerrain

tinmanjuggernaut:
This patch copies the raycast function from VoxelTerrain to VoxelLodTerrain, with appropriate updates to make it work. Please let me know if it's ok or needs adjustment.

The only issue with it is that it only raycasts on LOD 0. I experimented with other LODs and had mixed results from crashing on LOD 8, to not hitting on LOD 1. 

In game, it works as well as the raycasting on VoxelTerrain, once LOD 0 is loaded. So it's possible to fall through the local environment in the first few seconds, but once loaded, it's fine. 

Raycast collision makes a playable terrain until full physics can be developed. The LOD terrain is great so far and allows an older system like mine (Intel HD4400 laptop) to have a pretty far visual depth (1000) at a decent framerate (30-48fps even with short shadows).

~~If you want to see it in action, check out my branch here:
https://github.com/tinmanjuggernaut/voxelgame/tree/fps_demo~~
(Edit: this demo now uses physics based collision)

## Instructions
* Run heightmap.tscn to see raycast collision as is without this patch. 
* lod_noise.tscn needs this patch, but basically works the same. It uses 3D noise on the LOD terrain.
* This uses an input map defined in project.godot.
* Press 'F' to toggle between first and third person.
* W/A/S/D, space for jump or hold for jetpack. If you fall into the terrain, jetpack will push you up.

LOD-noise

### DIFF
--- a/terrain/voxel_lod_terrain.cpp
+++ b/terrain/voxel_lod_terrain.cpp
@@ -1,6 +1,7 @@
 #include "voxel_lod_terrain.h"
 #include "../math/rect3i.h"
 #include "../util/profiling_clock.h"
+#include "../util/voxel_raycast.h"
 #include "voxel_map.h"
 #include "voxel_mesh_updater.h"
 #include <core/engine.h>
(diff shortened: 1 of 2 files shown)


Zylann:
Sorry but the implementation has flaws, it only works with LOD0 and will fail if used further away.

tinmanjuggernaut:
I'm struggling on this one. Can you give me a clue? 

Here's what I see happening:
* _raycast_binding_predicate calls `map->get_voxel_f(pos)`
* which converts voxel to box coords 
* then requests that block `get_block(bpos)`
* which requests `_blocks.getptr(bpos)` which is a hashmap

I can iterate through all LODS (0-7) and call map->get_voxel_f(pos), but getptr returns null on all LODs until LOD 0 is displayed, even though 1-7 are on screen. 

How would you go about doing this?

Zylann:
I'm not sure yet regarding performance, but a correct way would be to check lod 0 if it has a wanted block. If it doesn't, check LOD1, and so on until none are available.

tinmanjuggernaut:
@Zylann, I and likely most people will probably use physics based collision for games, which means the physics based raycasting is already in place. It's far superior to the blocky, integer only raycasting from the terrain. I don't know what a terrain based collision would be useful for, except maybe mouse picking on blocky terrains when collision is turned off?

We know this patch is incomplete as it only works on lod0. I don't intend to make it work on all LODs since physics is better. So, I could update this with the current master... or I could just delete the branch and PR. What do you think?

Zylann:
I'm not planning to merge this one then, but I still have in mind to have that at some point because it's the only way to edit collision-less terrain.

PR #31, opened 2019-06-14
Emerge bugfix

### LINKED ISSUE #30: Default voxel stream immerge and emerge don't get called
The immerge_block and emerge_block do not seem to get called
I tried debugging with a print statement, but nothing happens or gets printed to the debugger or terminal window, leaving me with empty terrain.

tinmanjuggernaut:
Fixes two small bugs:

1. Voxel_name bindings were incorrect, which meant you couldn't read/write the variable like `print(voxel.voxel_name)`.

2. Calling emerge used the wrong overloaded version and resulted in a returned error, thus the call never got to the script.

fixes #30

### DIFF
--- a/streams/voxel_stream.cpp
+++ b/streams/voxel_stream.cpp
@@ -10,8 +10,15 @@ void VoxelStream::emerge_block(Ref<VoxelBuffer> out_buffer, Vector3i origin_in_v
 		Variant arg2 = origin_in_voxels.to_vec3();
 		Variant arg3 = lod;
 		const Variant *args[3] = { &arg1, &arg2, &arg3 };
-		//Variant::CallError err; // wut
-		script->call("emerge_block", args, 3);
+		Variant::CallError err;
+		script->call("emerge_block", args, 3, err);
+		if (err.error != Variant::CallError::CALL_OK) {
+			ERR_EXPLAIN(String("voxel_stream.cpp:emerge_block gave an error: ") + String::num(err.error)
+				+ String(" Argument: ") + String::num(err.argument)
+				+ String(" Expected type: ") + String(Variant::get_type_name(err.expected)));
+			ERR_FAIL();
+			// This had to be explicitely logged due to the usual GD debugger not working with threads
+		}
 	}
 }
 
@@ -24,8 +31,15 @@ void VoxelStream::immerge_block(Ref<VoxelBuffer> buffer, Vector3i origin_in_voxe
 		Variant arg2 = origin_in_voxels.to_vec3();
 		Variant arg3 = lod;
 		const Variant *args[3] = { &arg1, &arg2, &arg3 };
-		//Variant::CallError err; // wut
-		script->call("immerge_block", args, 3);
+		Variant::CallError err;
+		script->call("immerge_block", args, 3, err);
+		if (err.error != Variant::CallError::CALL_OK) {
+			ERR_EXPLAIN(String("voxel_stream.cpp:immerge_block gave an error: ") + String::num(err.error)
+				+ String(" Argument: ") + String::num(err.argument)
+				+ String(" Expected type: ") + String(Variant::get_type_name(err.expected)));
+			ERR_FAIL();
+			// This had to be explicitely logged due to the usual GD debugger not working with threads
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


streams/voxel_stream.cpp
```diff
@@ -10,8 +10,13 @@ void VoxelStream::emerge_block(Ref<VoxelBuffer> out_buffer, Vector3i origin_in_v
...
 		const Variant *args[3] = { &arg1, &arg2, &arg3 };
-		//Variant::CallError err; // wut
-		script->call("emerge_block", args, 3);
+		Variant::CallError err;
+		script->call("emerge_block", args, 3, err);
+		if (err.error != Variant::CallError::CALL_OK)
+			print_line(String("voxel_stream.cpp:emerge_block gave an error: ") + String::num(err.error)
+				+ String(" Argument: ") + String::num(err.argument)
+				+ String(" Expected type: ") + String(Variant::get_type_name(err.expected))
+			);
```
Zylann: I'm not sure if that's the correct way to report errors... from what I see elsewhere in the engine, calls are just made almost without error checking.

voxel.cpp
```diff
@@ -267,7 +267,7 @@ void Voxel::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("set_geometry_type", "type"), &Voxel::set_geometry_type);
 	ClassDB::bind_method(D_METHOD("get_geometry_type"), &Voxel::get_geometry_type);
 
-	ADD_PROPERTY(PropertyInfo(Variant::STRING, "voxel_name"), "set_name", "get_name");
+	ADD_PROPERTY(PropertyInfo(Variant::STRING, "voxel_name"), "set_voxel_name", "get_voxel_name");
```
Zylann: Can you separate this change to another PR?

streams/voxel_stream.cpp
Zylann: Due to the status of thread debugging, this could be a temporary solution to help reporting errors until it gets fixed in the engine. `print_error` may be better suited though?

tinmanjuggernaut: Ok, I don't know the correct way. I saw this in core/image.cpp which goes to error_macros.cpp which eventually does go to OS::get_singelton()->print_error):
```
ERR_EXPLAIN("Cannot convert to <-> from compressed formats. Use compress() and decompress() instead.");
ERR_FAIL();
```

I see other core files use this same format. Would you like me to call print_error directly, or use these ERR_EXPLAIN macros?

How's this?
```
		if (err.error != Variant::CallError::CALL_OK) {
			ERR_EXPLAIN(String("voxel_stream.cpp:emerge_block gave an error: ") + String::num(err.error)
				+ String(" Argument: ") + String::num(err.argument)
				+ String(" Expected type: ") + String(Variant::get_type_name(err.expected)));
			ERR_FAIL();
		}
```

voxel.cpp
tinmanjuggernaut: I will hold off separating unless you ask again.

streams/voxel_stream.cpp
tinmanjuggernaut: Or I could just cut out the error checking all together if desired. Just let me know.

Zylann: tbh I don't know what's correct here, from examples I saw I always though errors were reported automatically. So "usual way" would be to call using the error return param, but without print because Godot is supposed to do that already. But due to no debugger inside thread, having extra print would be useful as temporary workaround.

tinmanjuggernaut: So does that mean you want me to use the macros as I posted and push this to the PR as is? Do you want the voxel_name fix separated?

Zylann: You could go with those macros but leave a comment explaining this had to be explicitely logged due to the usual logger not working with threads.

tinmanjuggernaut: Actually, I didn't need to use that error check code that I put in. All the information I gave about the enum and the function tracing came from the debugger which worked just fine in VS2019. I didn't notice that I was on a separate thread. I just added it because I thought it was a good idea, but it could be removed.

Zylann: I was referring to the Godot debugger, not VS :)

tinmanjuggernaut:
Ok, changes have been committed.

PR #34, opened 2019-06-18
Fix compiling on gcc

### LINKED ISSUE #32: Won't compile with gcc
master
gcc 8.3

I'm moving my dev environment to linux. Gcc won't compile Voxel tools out of the box, but will with the following changes. There are two things to fix, and a whole bunch of warnings. @Zylann, I have some questions for you on how you want to fix things before I submit a PR.

**1. compute_lod_count returns an int but is defined as void.**
Easy. Diff below.

**2. cannot bind non-const lvalue reference of type 'LodOctree<bool>::NoDestroyAction&' to an rvalue of type 'LodOctree<bool>::NoDestroyAction'**

```modules/voxel/terrain/voxel_lod_terrain.cpp: In member function 'void VoxelLodTerrain::set_lod_count(unsigned int)':
modules/voxel/terrain/voxel_lod_terrain.cpp:158:71: error: cannot bind non-const lvalue reference of type 'LodOctree<bool>::NoDestroyAction&' to an rvalue of type 'LodOctree<bool>::NoDestroyAction'
   _lod_octree.create_from_lod_count(bs, p_lod_count, LodOctree<bool>::NoDestroyAction());
```

[This post](https://stackoverflow.com/questions/27172033/invalid-initialization-of-non-const-reference-of-type) says it works fine in MSVC but is not portable. They said it could be fixed by making the referenced argument const. That of course made a whole cascade of other errors, which were fixed by making the operator() itself const. 

I don't know if this is correct, but the changes allowed it to compile, run and my own LOD+raycast demo worked as expected.

However, I noted there were other functions that used the same style of passing "A& destroy_action" such as update(), clear(), etc. Should they have a const argument also? What do you want to do here?

Here is the minimum needed to compile:

```
diff --git a/terrain/lod_octree.h b/terrain/lod_octree.h
index 4ddc646..939d426 100644
--- a/terrain/lod_octree.h
+++ b/terrain/lod_octree.h
@@ -37,7 +37,7 @@ public:
        };
 
        struct NoDestroyAction {
-               inline void operator()(Node *node, int lod) {}
+               inline void operator()(Node *node, int lod) const {}
        };
 
        template <typename A>
@@ -47,7 +47,7 @@ public:
                _base_size = 0;
        }
 
-       static void compute_lod_count(int base_size, int full_size) {
+       static int compute_lod_count(int base_size, int full_size) {
                int po = 0;
                while (full_size > base_size) {
                        full_size = full_size >> 1;
@@ -57,7 +57,7 @@ public:
        }
 
        template <typename A>
-       void create_from_lod_count(int base_size, unsigned int lod_count, A &destroy_action) {
+       void create_from_lod_count(int base_size, unsigned int lod_count, const A &destroy_action) {
                ERR_FAIL_COND(lod_count > 32);
                clear(destroy_action);
                _base_size = base_size;
diff --git a/terrain/voxel_lod_terrain.h b/terrain/voxel_lod_terrain.h
index 7ec903b..aca8c60 100644
--- a/terrain/voxel_lod_terrain.h
+++ b/terrain/voxel_lod_terrain.h
@@ -78,7 +78,7 @@ private:
        bool check_block_loaded_and_updated(const Vector3i &p_bpos, unsigned int lod_index);
 
        template <typename A>
-       void for_all_blocks(A &action) {
+       void for_all_blocks(const A &action) {
                for (int lod_index = 0; lod_index < MAX_LOD; ++lod_index) {
                        if (_lods[lod_index].map.is_valid()) {
                                _lods[lod_index].map->for_all_blocks(action);
```

**3. There are a ton of warnings from the voxel module.**
So many that I had to turn off warnings just so I could see the errors above to fix them. They are only warnings, but should probably be fixed at some point. I can fix some of them now if you like, but some are interwoven throughout the whole module. 

Easy fix: 
```
modules/voxel/terrain/voxel_map.h:96:3: warning: suggest parentheses around assignment used as truth value [-Wparentheses]
   while (key = _blocks.next(key)) {
```

Most have to do with comparing a signed int to an unsigned int. Some are an easy fix:

```
modules/voxel/terrain/block_thread_manager.h:172:37: warning: comparison of integer expressions of different signedness: 'int' and 'unsigned int' [-Wsign-compare]
   for (int job_index = 0; job_index < _job_count && i < input.blocks.size(); ++job_index) {
```

Some have a caveat:
```
_FORCE_INLINE_ bool validate_pos(unsigned int x, unsigned int y, unsigned int z) const {
                return x < _size.x && y < _size.y && z < _size.z;
        }
```
_size is a Vector3i which uses signed ints. So this function is easy to fix. But then you have similar functions that do not compare and do not give warnings, but they still use unsigned ints when the underlying data are signed. Is this a problem?

```
_FORCE_INLINE_ unsigned int get_volume() const {
                return _size.x * _size.y * _size.z;
        }
```

Some sign comparison complaints are in code that looks like this. I don't want to just go through and start changing functions around. Do I change the arguments in hermite_value, or the return type in voxel_buffer? Perhaps it makes sense to change _size to an unsigned vector3i and make everything unsigned ints?
```
meshers/dmc/hermite_value.h:
inline float get_isolevel_clamped(const VoxelBuffer &voxels, unsigned int x, unsigned int y, unsigned int z) {
        x = x >= voxels.get_size().x ? voxels.get_size().x - 1 : x;


voxel_buffer.h:
_FORCE_INLINE_ int get_size_x() const { return _size.x; }
```

What would you like to do about these warnings?

tinmanjuggernaut:
Not complete. I'll look at addressing the final warning tomorrow.

sign comparison warnings is the biggest block. Commits separated for sanity.

fixes #32


terrain/voxel_lod_terrain.cpp
```diff
@@ -575,7 +585,7 @@ void VoxelLodTerrain::_process() {
 
 		//print_line(String("Loaded {0} blocks").format(varray(output.emerged_blocks.size())));
 
-		for (int i = 0; i < output.blocks.size(); ++i) {
+		for (unsigned int i = 0; i < (unsigned int)output.blocks.size(); ++i) {
```
Zylann: this one didnt need to be int

terrain/voxel_lod_terrain.cpp
```diff
@@ -686,7 +696,7 @@ void VoxelLodTerrain::_process() {
 			_block_updater->pop(output);
 			_stats.updater = output.stats;
 
-			for (unsigned int i = 0; i < output.blocks.size(); ++i) {
+			for (unsigned int i = 0; i < (unsigned int)output.blocks.size(); ++i) {
```
Zylann: `i` didnt need to be unsigned

terrain/voxel_lod_terrain.cpp
```diff
@@ -742,9 +752,9 @@ void VoxelLodTerrain::_process() {
 			Ref<ArrayMesh> mesh;
 			mesh.instance();
 
-			int surface_index = 0;
+			unsigned int surface_index = 0;
 			const VoxelMeshUpdater::OutputBlockData &data = ob.data;
-			for (int i = 0; i < data.smooth_surfaces.surfaces.size(); ++i) {
+			for (unsigned int i = 0; i < (unsigned int)data.smooth_surfaces.surfaces.size(); ++i) {
```
Zylann: Didn't need to be unsigned

voxel_buffer.h
```diff
@@ -86,7 +86,7 @@ class VoxelBuffer : public Reference {
 	void copy_from(const VoxelBuffer &other, unsigned int channel_index = 0);
 	void copy_from(const VoxelBuffer &other, Vector3i src_min, Vector3i src_max, Vector3i dst_min, unsigned int channel_index = 0);
 
-	_FORCE_INLINE_ bool validate_pos(unsigned int x, unsigned int y, unsigned int z) const {
+	_FORCE_INLINE_ bool validate_pos(int x, int y, int z) const {
 		return x < _size.x && y < _size.y && z < _size.z;
```
Zylann: _size was the one to cast instead

voxel_isosurface_tool.cpp
```diff
@@ -33,7 +33,7 @@ namespace {
 
 inline void do_op(VoxelBuffer &buffer, int x, int y, int z, float d1, VoxelIsoSurfaceTool::Operation op) {
 
-	float res;
+	float res=0;
```
Zylann: Does GCC still complain if instead of giving a default value, you add a `default` label in the switch?

terrain/voxel_lod_terrain.cpp
tinmanjuggernaut: ok, reverted

terrain/voxel_lod_terrain.cpp
tinmanjuggernaut: i was originally unsigned, which produces the warning. I'll remove unsigned from both so it matches above.

terrain/voxel_lod_terrain.cpp
tinmanjuggernaut: ok, removed

voxel_buffer.h
tinmanjuggernaut: Ok, I have cast each of the _size.x's to unsigned.

voxel_isosurface_tool.cpp
tinmanjuggernaut: Yes. The default section didn't affect the warnings, which come later on:

We get four of these lovely warnings:

```
In file included from modules/voxel/voxel_isosurface_tool.h:4,
                 from modules/voxel/voxel_isosurface_tool.cpp:1:
modules/voxel/voxel_buffer.h: In member function 'void VoxelIsoSurfaceTool::do_heightmap(Ref<Image>, Vector3, real_t, VoxelIsoSurfaceTool::Operation)':
modules/voxel/voxel_buffer.h:41:34: warning: 'res' may be used uninitialized in this function [-Wmaybe-uninitialized]
   int v = static_cast<int>(128.f * iso + 128.f);
                            ~~~~~~^~~~~
modules/voxel/voxel_isosurface_tool.cpp:36:8: note: 'res' was declared here
  float res;
        ^~~
```

tinmanjuggernaut:
Ok, these last commits revert your noted changes and removes the final warning. Please review and let me know what you'd like to change.

voxel_isosurface_tool.cpp
Zylann: the thing is, GCC is completely wrong about that :p `res` would be initialized in all `case` labels, AND the `default` anyways. A remaining case is if an exception is thrown, which is turned off in Godot. It worries me a little because this function can run a very high number of times.

I guess you can initialize for now, tbh even the `MIN` and `MAX` macros are bad for perf here, because they expand into more function calls than necessary :p

tinmanjuggernaut: I was confused about what I was supposed to test. The warning remained with an empty default. However if I initialize to 0 in default case, the warning goes away. PR updated

PR #35, opened 2019-07-03
Allow VoxelTerrain.set_material() from gdscript

tinmanjuggernaut:
Currently VoxelTerrain doesn't bind set_material, so you can't set a material from code like you can with  VoxelLodTerrain. This PR adds the bindings.

### DIFF
--- a/terrain/voxel_terrain.cpp
+++ b/terrain/voxel_terrain.cpp
@@ -963,6 +963,9 @@ void VoxelTerrain::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("set_voxel_library", "library"), &VoxelTerrain::set_voxel_library);
 	ClassDB::bind_method(D_METHOD("get_voxel_library"), &VoxelTerrain::get_voxel_library);
 
+	ClassDB::bind_method(D_METHOD("set_material", "id", "material"), &VoxelTerrain::set_material);
+	ClassDB::bind_method(D_METHOD("get_material", "id"), &VoxelTerrain::get_material);
+
 	ClassDB::bind_method(D_METHOD("set_view_distance", "distance_in_voxels"), &VoxelTerrain::set_view_distance);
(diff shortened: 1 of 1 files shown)


Zylann:
Problem: if you use `set_material` from script, currently that won't change the existing blocks mesh material. It's one of many other parameters I haven't focused on exposing because if they need to change at runtime there is a lot of housekeeping to do / regenerate the entire map.
This one might be simple though, by iterating through all the blocks (expect a stutter if you do).

tinmanjuggernaut:
Oh, I see that I can do `set("materials/0/", mymaterial)`. That's a little obscure, so all this patch does is makes that a little more obvious to people who would want to create it from code (such as if they are creating a custom stream).

The problem you cite is there either way, but this patch certainly doesn't hurt anything. If needed they could delete and recreate the whole terrain, such as loading a different level.

Zylann:
Yeah the reason you can use the `set` is because that's how they get loaded and saved.
If people want  to create the terrain from code they should setup everything before adding to tree though. Once in the tree the terrain will start its jobs which might miss the change.

tinmanjuggernaut:
When I run Godot, everything works, but I now get this message:
```
ERROR: add_property: Invalid Function for Setter: VoxelTerrain::set_material for property: material
   At: core/class_db.cpp:933.
```
I had copied this line from VoxelLODTerrain.cpp:
```
ADD_PROPERTY(PropertyInfo(Variant::OBJECT, "material", PROPERTY_HINT_RESOURCE_TYPE, "Material"), "set_material", "get_material");
```

Apparently, that was wrong and Godot doesn't like it. Is it because set_material/get_material require an index?  If I comment it out of VoxelTerrain, the error goes away and I can still set_material from either code or the editor. All the other demos seem to work fine.

I've fixed it on my branch. Should I submit a PR to revert, then another to fix, or just the fix PR?

Zylann:
Oh wait, indeed the property part is wrong, you should remove it.

PR #44, opened 2019-08-10
Added info in order to enable AO

### LINKED ISSUE #43: Add on docs to enable vertex color in order to enable AO
Hello folks!

Maybe I missed something, but I discovered by reading the source that we got a nice Ambient Occlusion if we just enable Vertex Color (Use as Albedo) on Material. When I first read the getting started, this wasn't obvious for me, so I think this should have a better emphasis on docs.

Thanks!

afonsolage:
Added on docs how to enable AO, as mentioned on #43

fixes #43

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ A C++ module for creating volumetric worlds in Godot Engine.
 Features
 ---------------------------
 
-- Realtime editable, 3D based terrain (Unlike a hightmap based terrain, this allows for overhangs, tunnels, and user creation/destruction)
+- Realtime editable, 3D based terrain (Unlike a heightmap based terrain, this allows for overhangs, tunnels, and user creation/destruction)
 - Full collision support
 - Infinite terrains made by paging sections in and out
 - Voxel data is streamed from a variety of sources, which includes the ability to write your own
--- a/doc/04_materials.md
+++ b/doc/04_materials.md
@@ -38,6 +38,7 @@ Rather than writing your own shader from scratch, especially with triplanar mapp
 
 If you want to start without a texture and just want to use a color, try turning down roughness, or adding some metalic to give the surface some reflectivity. This will allow light to reflect off the curves of the terrain in the distance. Otherwise you'll just see an undifferentiated mass of color.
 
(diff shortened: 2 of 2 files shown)


Zylann:
I think you should precise that it's to be able to *see* the result of baked AO. Enabling vertex color does not *activate* AO, it's already there if the option is active (not exposed yet but one day it will). Also this kind of AO only exists for blocky voxels.

tinmanjuggernaut:
The PR was done right.

@Zylann 
So you mean VT puts AO into the vertex color? So basically like blender's dirty vertex painting? And enabling vertex_color_use_as_albedo in the material mixes the vertex color with the albedo. Right?

In that case, @afonsolage, I'd write something like this:

```
Also, VoxelTerrain adds Ambient Occlusion to the vertex colors (blocky only). You can add this to your material by enabling the `Vertex Color/Use As Albedo` flag in your material.
```

Please also update Readme.md and change `Unlike a hightmap based terrain` to `Unlike a heightmap based terrain`.

These changes can then be squashed and force uploaded to your fork (see the doc I sent you). The PR will automatically update.
Thanks!

Zylann:
> So you mean VT puts AO into the vertex color? So basically like blender's dirty vertex painting? And enabling vertex_color_use_as_albedo in the material mixes the vertex color with the albedo. Right?

Yes, that's correct :)

tinmanjuggernaut:
Also you should put in " fixes #43 " in your initial post and it will automatically close that issue once the final PR is accepted.

afonsolage:
Done! What about now?

tinmanjuggernaut:
Good. Thanks for taking care of my typo. Finally, you'll want to squash your changes into one commit.  That doc I sent you shows you how, but basically you have three commits so you do this on your branch:
`git rebase -i HEAD~3`

Then keep the first commit and change the others to `fixup`. It will squash the extra two commits into the first and erase their commit messages (`squash` will keep the commit messages).
Then `git push --force origin update_material_docs` to overwrite your rebased log on your repository.

afonsolage:
I had to change to `git rebase -i @~3` because i'm running on linux (base on [this answer](https://stackoverflow.com/a/47233396)). But anyway, This is my first PR, as you can see, so I'm learning a lot :+1:

PR #74, opened 2019-10-03
Allow building against Godot 3.1 branch

### LINKED ISSUE #64: Compile error on Windows 10 with Godot 3.1
I am unable to compile Godot engine with voxel module as I am getting following error:

```
[Initial build] Compiling ==> modules\voxel\streams\voxel_block_serializer.cpp
voxel_block_serializer.cpp
modules\voxel\streams\voxel_block_serializer.cpp(90): error C3861: 'ERR_FAIL_COND_V_MSG': identifier not found
modules\voxel\streams\voxel_block_serializer.cpp(119): error C3861: 'ERR_FAIL_COND_V_MSG': identifier not found
modules\voxel\streams\voxel_block_serializer.cpp(164): error C3861: 'ERR_FAIL_COND_V_MSG': identifier not found
modules\voxel\streams\voxel_block_serializer.cpp(167): error C3861: 'ERR_FAIL_COND_V_MSG': identifier not found
scons: *** [modules\voxel\streams\voxel_block_serializer.windows.opt.tools.64.obj] Error 2
scons: building terminated because of errors.
```

I am compiling on Windows 10 Pro 64bit. I tried to compile using MinGW and MSVC with same result:
```
scons platform=windows target=release_debug
scons platform=windows target=release_debug use_mingw=yes
```

I have no problem to compile Godot without voxel module.

tinmanjuggernaut:
Tested on linux/gcc w/ and w/o mono.

Fixes #64

### DIFF
--- a/error_macros.h
+++ b/error_macros.h
@@ -1,0 +1,22 @@
+// Macros defined in Godot 3.2 in core/error_macros.h
+// Copied here for Voxel Tools 3.1 branch
+
+#define ERR_FAIL_COND_MSG(m_cond, m_msg)                                                                   \
+	{                                                                                                      \
+		if (unlikely(m_cond)) {                                                                            \
+			ERR_EXPLAIN(m_msg);                                                                            \
+			_err_print_error(FUNCTION_STR, __FILE__, __LINE__, "Condition ' " _STR(m_cond) " ' is true."); \
+			return;                                                                                        \
+		}                                                                                                  \
+		_err_error_exists = false;                                                                         \
+	}
+
+#define ERR_FAIL_COND_V_MSG(m_cond, m_retval, m_msg)                                                                                 \
+	{                                                                                                                                \
+		if (unlikely(m_cond)) {                                                                                                      \
+			ERR_EXPLAIN(m_msg);                                                                                                      \
+			_err_print_error(FUNCTION_STR, __FILE__, __LINE__, "Condition ' " _STR(m_cond) " ' is true. returned: " _STR(m_retval)); \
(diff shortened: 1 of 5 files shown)


Zylann:
You added `error_macros.h`, but aren't these already defined if you compile using Godot 3.1?

tinmanjuggernaut:
Those macros are not in 3.1 and it will not build without them.

> ERR_FAIL_COND_V_MSG is a Godot macro which looks like it was added last month.
> https://github.com/godotengine/godot/blame/master/core/error_macros.h

Zylann:
Yeah sorry for some reason I thought you did the opposite...

tinmanjuggernaut:
> tbh we could have wrapped this into an `#ifndef ERR_FAIL_COND_MSG` without needing to branch, but that branch at least makes it obvious that this compiles for that version.

Your call. This way we'll have to backport and maintain at least two branches, stable and master.

It wasn't just the error macros. It was also one function call. A simple fix, but over time the godot code base will diverge more between versions. Especially with vulkan and improving VT with the visual server to fix the vsync isses, etc. 

Branches is probably cleaner. It's easy to switch back and forth just by checking out the godot and VT branches.

blockspacer:
@tinmanjuggernaut 
Thanks for contribution.
Can someone explain why  `error_macros.h` without `#pragma once` or `#ifndef HEADER_H ... `?

tinmanjuggernaut:
Poor programming practice on my part. However it's just two defines, so if it gets preprocessed multiple times it will waste exactly 440ns of your life. :)

PR #76, opened 2019-10-04
Update docs

tinmanjuggernaut:


doc/03_create-terrain.md
```diff
@@ -16,7 +16,48 @@ Now that your Godot Engine has voxel support built in, you can either download t
...
 1. Play your scene and you should see a terrain.
 
-![Generated voxel terrain](images/default-terrain.jpg)
+
+<img src="images/default-terrain.jpg" width="800" />
+
+
+
+## Create A Terrain In The Editor: VoxelLODTerrain + 3D Noise
+Here we'll look at VLT with a noise data stream.
```
Zylann: Don't use initials, prefer writing the full name of the node within backticks.

doc/07_performance-tips.md
```diff
@@ -0,0 +1,16 @@
+# Performance Tips
+
+* Disable VSync and set threading to Single Safe in the project settings. Terrain building and editing performs significnatly faster with these settings.
+
+* Excessive noise height will kill your performance, use with care. A height range of 300 can run at 300-400fps on a GTX1060, but a range of 6-10,000 might run at 30-40fps. Hopefully this will be addressed when occlusion culling is available in Godot 4.
```
Zylann: It may also be caused by the fact invisible LODs are still included in Godot's frustrum culling, which is itself not very good compared to what's coming in 4.0.

doc/07_performance-tips.md
```diff
@@ -0,0 +1,16 @@
...
+
+* Lower shadow distance. I might use 200-400 on a GTX1060, but maybe 20 or disabled on an integrated card.
```
Zylann: I wish I could tell the larger shadow maps to use only the lower LODs, but godot doesn't have any way to customize how shadows render so they always render the same amount of meshes the player would see from its point of view :shrug:

tinmanjuggernaut:
updated

Zylann:
Thanks!

Issue #110, opened 2020-01-20
No voxels showing on linux


creikey:
After building  off of the latest godot master with the module in the `modules/` folder, I was able to successfully open [the demo](https://github.com/Zylann/voxelgame) , however there are no voxels visible, and the main screen looks as following:
![image](https://user-images.githubusercontent.com/28424206/72691990-44d23900-3ade-11ea-970f-53711412e748.png)

There are no errors in console or editor.

creikey:
Does not work on GLES2 either

tinmanjuggernaut:
Did you change your channel to sdf?

blockspacer:
@creikey
Can you see voxels in https://github.com/tinmanjuggernaut/voxelgame ?

creikey:
@blockspacer No, I cannot.

@tinmanjuggernaut Sorry, I'm unsure of what you mean by channel in this context, could you clarify?

tinmanjuggernaut:
See #4 https://github.com/tinmanjuggernaut/godot_voxel/blob/master/doc/03_create-terrain.md
Open up your stream settings and try changing the channel.

Also there are several demos in Zylann's and in mine, both linked above. You can't see voxels in any of them? It's been working for a while cross platform, so it's more likely you have a wrong setting (Zylann's demo may have it wrong by default as some things recently changed in the noise generator).

If none of these work you need to post more details and do more thorough testing. One sentence doesn't give us any information with which to help you.

blockspacer:
@creikey 

Can you provide more details about your hardware (CPU, GPU), please.

Can you provide commit hashes of https://github.com/godotengine/godot and https://github.com/tinmanjuggernaut/voxelgame ?

Can you run other 3d projects with godot ?

BTW: godot_voxel HTML5 builds are not supported, produces similar result

tinmanjuggernaut:
I just built Godot and VT master for linux. All my demos (in /fps_demo/) work. Your screenshot is from one of Zylann's demos. Most of which work, but some don't. You can also try my binaries: http://tokisan.com/godot-binaries

Zylann:
I made a few breaking changes recently, some symbols got renamed. My demo is fairly outdated at this point.
Check if you get any errors in the log.

tobischw:
I am running into the same issue on Windows. I am following [this](https://github.com/Zylann/godot_voxel/blob/master/doc/03_create-terrain.md) tutorial. 

[...]

PR #118, opened 2020-02-12
VT/VLT Detect channel

tinmanjuggernaut:
Changes:
* Channel set by generators
* Moves channel up to the first parameter for all end user generators (added manually in each child class so it shows up first :/ )
* Default channel is SDF (since VT can do either and VLT only does SDF)
* Both VT and VLT detect channel from generators
	* VT uses either channel
	* VT/VLT turns on an editor warning if generator uses an unsupported channel
	* Sets VoxelTool channel
* Fix doc tools
* Updated written and generated documentation
* ~~Limits Stream dropdown list (in Inspector only) to only usable streams and generators~~

Tested on windows and linux.


streams/voxel_stream.cpp
```diff
@@ -5,6 +5,16 @@
...
 
+void VoxelStream::set_channel(VoxelBuffer::ChannelId channel) {
+	ERR_FAIL_INDEX(channel, VoxelBuffer::MAX_CHANNELS);
+	_channel = channel;
+	emit_changed();
+}
+
+VoxelBuffer::ChannelId VoxelStream::get_channel() const {
+	return _channel;
+}
```
Zylann: What is the reasoning behind having a `channel` property in `VoxelStream`? At this level of abstraction, a stream can provide 1, 2, or all 8 channels at once, from either procedural sources or existing files.

streams/voxel_stream_file.cpp
```diff
@@ -79,6 +79,6 @@ void VoxelStreamFile::_bind_methods() {
 
 	ClassDB::bind_method(D_METHOD("get_block_size"), &VoxelStreamFile::_get_block_size);
 
-	ADD_PROPERTY(PropertyInfo(Variant::OBJECT, "fallback_stream", PROPERTY_HINT_RESOURCE_TYPE, "VoxelStream"), "set_fallback_stream", "get_fallback_stream");
+	ADD_PROPERTY(PropertyInfo(Variant::OBJECT, "fallback_stream", PROPERTY_HINT_RESOURCE_TYPE, "VoxelGeneratorImage,VoxelGeneratorNoise,VoxelGeneratorNoise2D,VoxelGeneratorTest,VoxelStreamBlockFiles,VoxelStreamRegionFiles"), "set_fallback_stream", "get_fallback_stream");
```
Zylann: Not having abstract classes such as `VoxelStream` or `VoxelGenerator` prevents the implementation of a custom one using a script. `VoxelStreamFile` is the abstract class for file-based streams.

streams/voxel_stream_region_files.cpp
```diff
@@ -1141,6 +1141,7 @@ void VoxelStreamRegionFiles::_bind_methods() {
 
 	ClassDB::bind_method(D_METHOD("convert_files", "new_settings"), &VoxelStreamRegionFiles::convert_files);
 
+	ADD_PROPERTY(PropertyInfo(Variant::INT, "channel", PROPERTY_HINT_ENUM, VoxelBuffer::CHANNEL_ID_HINT_STRING), "set_channel", "get_channel");
```
Zylann: In the case of files, it's again a bit strange to have channel settings, especially a single one. That choice may happen earlier than that, in the generator. But maybe there is another intention or use case behind this?

terrain/voxel_lod_terrain.cpp
```diff
@@ -83,6 +83,13 @@ VoxelLodTerrain::~VoxelLodTerrain() {
 	}
 }
 
+String VoxelLodTerrain::get_configuration_warning() const {
```
Zylann: Warning when a part of the data will not get represented is interesting, but the design of how this information gets detected would probably need more thought

streams/voxel_stream.cpp
tinmanjuggernaut: I don't know? I started with it in voxel generator, but then my gdscript terrain demo wouldn't work because I couldn't inherit from voxelgenerator and get it to generate anything.

Also since you had previously said the streams (before there were generators) should dictate the channel, not the VT nodes,  I figured that every stream/generator should have the variable. 

Tbh, I don't know what the streams are for if you're going to separate the generators from them.

streams/voxel_stream_file.cpp
tinmanjuggernaut: My gdscript demo extends VoxelStream with this configuration. This just removes it from the drop down list in the inspector. It doesn't affect the code, AFAIK and have seen.

streams/voxel_stream_region_files.cpp
tinmanjuggernaut: No intention in the files. Just fumbling around to make the user interface simpler. I just wanted to remove the smooth meshing variable, and have the terrain nodes detect the channel type and it turned into a huge project.

Zylann:
> Default channel is SDF (since VT can do either and VLT only does SDF)

I'm still not decided wether or not removing smooth support from `VoxelTerrain`. It's not much code though.

> I'm not really sure about the voxelstreamfiles so didn't test them. But they don't seem to set any voxels so I'm not sure how they are used or how they create voxels.

`VoxelStreamFiles` is the base class for file-based streams, for which the logic is generally "load a file, or load from a fallback stream (generator) if not found".

> You have a note about wanting generators to not inherit from streams. But generators have a generate_block that pass a reference, so no way to access that via gdscript.

Voxel generators have a similar API to `emerge_block`, I don't understand what the problem is? 
They inherit from streams for now because I didnt want to break things too much. I feel like there may be some redesign to do in the future to make a clear separation between the *persistence of voxels* and what serves as *base generation* of the terrain.

About the VoxelLibrary crash, it's a bug, you could create an issue to track it.

streams/voxel_stream.cpp
tinmanjuggernaut: I could move the functions back to the generators and ignore streams entirely.

But that generate block function passes by reference. I didn't think that could be used in gdscript? And I had other issues with gdscript generation as noted.

streams/voxel_stream_file.cpp
Zylann: But then it means you need to set it via code? Or maybe you had it already in the property?

streams/voxel_stream.cpp
Zylann: `emerge_block` passed by reference too, I don't understand what the problem is

streams/voxel_stream_file.cpp
tinmanjuggernaut: Setting it in gdscript is what I mean. I don't see how there's any way to use the abstract classes in the inspector since they have no options or implementation. That's all that was removed: Useless options from the inspector.

In gdscript they can still be used, and it seems to be the only way to use them:

Extends VoxelStream:
https://github.com/tinmanjuggernaut/voxelgame/blob/master/project/fps_demo/scripts/MyStream.gd

Loads terrain based on custom VoxelStream:
https://github.com/tinmanjuggernaut/voxelgame/blob/master/project/fps_demo/scripts/CodeTerrain.gd

streams/voxel_stream.cpp
tinmanjuggernaut: Ok, maybe I just imagined the problem. I'll move the channel functions to voxel generator and ignore the streams. Then I'll see if I can get the gdscript working. Everything else worked but that. Maybe I just need a break from it.

terrain/voxel_lod_terrain.cpp
tinmanjuggernaut: You can see the only current use case that generates a warning is VLT + Type which results in a blank screen for the user. That's likely the cause behind #110 and has certainly caused me and others a lot of headaches.

Zylann: I had that too when doing blocky stuff at some point so it's a good idea, just needs different source for the info

streams/voxel_stream.cpp
tinmanjuggernaut: Channel functions are now only in the generators.

Creating a terrain based in Gdscript works fine. I have a [branch](https://github.com/tinmanjuggernaut/voxelgame/tree/detect_channel) in my demo that works.

Incidentally, none of your demos work with the changes you've been making recently.

terrain/voxel_lod_terrain.cpp
tinmanjuggernaut: Now it only queries when they are generators.

streams/voxel_stream_file.cpp
tinmanjuggernaut: Are you still concerned about this? With my current patches, the abstract classes are not visible in the inspector dropdown, but they are still accessible via gdscript and show up in the help. My demo extends VoxelGenerator just fine and it extended VoxelStream before that without issue.

tinmanjuggernaut:
It works fine (and much faster) on linux. Though there are build warnings (probably from the many uppdates). I'll look at those and also work on updating the docs.

tinmanjuggernaut:
Ok, written and generated documentation has been updated, plus some fixes to the doctool script. I reverted all changes to the streams, except for the limited list of fallback streams (for the inspector only). All my demos work (in my pending branch), including all the gdscript options. And I don't think any of your demos work.

terrain/voxel_lod_terrain.cpp
Zylann: I think it's better if `VoxelStream` has a function to get a mask of used channels, rather than limiting it to a single channel on generators only.

streams/voxel_stream_file.cpp
Zylann: If you make a custom script (which will them inherit an abstract class) and want to assign it in the inspector, it is a valid option to create it on the property and assign the script, or even drag and drop an existing one.

tinmanjuggernaut: I made one and tried it out. Ok, I see. I have reverted these changes.

terrain/voxel_lod_terrain.cpp
tinmanjuggernaut: `VoxelStream` now has nothing to do with channels. That was moved to `VoxelGenerator`. Do you want to keep it like this?

Instead of get_channel returning an enum, you want it to return an 8-bit mask?

Then I could have the enum values be 0x1, 0x2, 0x4, 0x8 instead of 1, 2, 3, 4. Then the bits and enum will still work together.  Then the VLT check still remains essentially the same as it warns on all channels except for SDF at the moment. I'll start working on this.

Zylann: Channel values can remain the same, they only turn as bitmask here. Conversion is straightforward, `1 << channel` to obtain mask from single channel.

tinmanjuggernaut:
I've updated this PR. It includes `get_used_channels_mask` in VoxelStream and tests for it in `VoxelTerrain`. However, this implementation has a problem.

VoxelTerrain uses this same function to set the channel for do_sphere, etc. But when using a gdscript generator, it won't call the gscript `get_used_channels_mask`. It calls VoxelStream::get_used_channels_mask, so it doesn't use the right channel.

It worked fine before when VoxelGenerator stored the channel, and gdscript inherited from that. 


Also note, I've put the ADDPROPERTY channel into the Image/Noise2D child classes, though the variable is in VGheightmap.cpp.  My aim was to have it be the first property in the list, as it can get lost when the noise settings are expanded (esp w/ fastnoise).

Zylann:
> when using a gdscript generator, it won't call the gscript get_used_channels_mask

That function should be implementable by GDScript, the same way emerge/immerge is.

> Also note, I've put the ADDPROPERTY channel into the Image/Noise2D child classes, though the variable is in VGheightmap.cpp

That's quite unfortunate... changing an API just because of editor display problems xD

doc/08_api-overview.md
```diff
@@ -167,26 +162,26 @@ The underlying storage for voxels. Also used to pass voxels between functions.
...
+### Voxel (blocky)
+A basic voxel unit. Creating Voxels with VoxelLibrary is recommended.
 
 * Stores ID, name, material, transparency and type
 
 
-### VoxelLibrary
-A factory for creating Voxels. (blocky only)
+### VoxelLibrary (blocky)
+A factory for creating Voxels.
```
Zylann: A collection of voxel types, which is used for blocky only, similar to Minecraft or GridMap.

doc/tools/build.py
```diff
@@ -118,7 +118,7 @@ def find_godot(godot, bindir):
     else:
         filemask = 'godot.*.[36][24]'
 
-    binfiles = glob.glob(bindir+filemask)
+    binfiles = glob.glob(str(bindir/filemask))
```
Zylann: `os.path.join(bindir, filemask)`?

generators/voxel_generator_noise.cpp
```diff
@@ -1,16 +1,21 @@
 #include "voxel_generator_noise.h"
 
+void VoxelGeneratorNoise::set_noise(Ref<OpenSimplexNoise> noise) {
+	_noise = noise;
+}
+
 void VoxelGeneratorNoise::set_channel(VoxelBuffer::ChannelId channel) {
 	ERR_FAIL_INDEX(channel, VoxelBuffer::MAX_CHANNELS);
 	_channel = channel;
+	emit_changed();
```
Zylann: Should emit only if `_channel` actually change

generators/voxel_generator_heightmap.cpp
```diff
@@ -8,12 +8,17 @@ VoxelGeneratorHeightmap::VoxelGeneratorHeightmap() {
 void VoxelGeneratorHeightmap::set_channel(VoxelBuffer::ChannelId channel) {
 	ERR_FAIL_INDEX(channel, VoxelBuffer::MAX_CHANNELS);
 	_channel = channel;
+	emit_changed();
```
Zylann: Should emit only if `_channel` actually changed

generators/voxel_generator_noise.cpp
```diff
@@ -1,16 +1,21 @@
 #include "voxel_generator_noise.h"
 
+void VoxelGeneratorNoise::set_noise(Ref<OpenSimplexNoise> noise) {
+	_noise = noise;
```
Zylann: Technically it should emit `changed` for any property modification, although right now nothing uses that

generators/voxel_generator_test.cpp
```diff
@@ -101,7 +115,7 @@ void VoxelGeneratorTest::generate_block_waves(VoxelBuffer &out_buffer, Vector3i
 					rh = size.y;
 
 				for (int ry = 0; ry < rh; ++ry) {
-					out_buffer.set_voxel(_voxel_type, rx, ry, rz, 0);
+					out_buffer.set_voxel(_voxel_type, rx, ry, rz, _channel);
```
Zylann: For SDF this is not correct

terrain/voxel_lod_terrain.h
```diff
@@ -22,6 +22,8 @@ class VoxelLodTerrain : public Spatial {
 	VoxelLodTerrain();
 	~VoxelLodTerrain();
 
+	String get_configuration_warning() const;
```
Zylann: Add `override` after `const`

terrain/voxel_terrain.cpp
```diff
@@ -27,7 +28,10 @@ VoxelTerrain::VoxelTerrain() {
 	_block_updater = NULL;
 
 	_run_in_editor = false;
-	_smooth_meshing_enabled = false;
+
+	Ref<VoxelLibrary> library;
+	library.instance();
+	set_voxel_library(library);
```
Zylann: Note that following this logic, we could just as well ask Godot to have a tileset by default in `TileMap` and a default MeshLibrary as well in `GridMap`

terrain/voxel_terrain.cpp
```diff
@@ -384,7 +375,15 @@ void VoxelTerrain::start_updater() {
 
 	// TODO Thread-safe way to change those parameters
 	VoxelMeshUpdater::MeshingParams params;
-	params.smooth_surface = _smooth_meshing_enabled;
+
+	if (_stream.is_valid()) {
+		if (_stream->get_used_channels_mask() & 1<<VoxelBuffer::CHANNEL_TYPE) {
+			params.smooth_surface = false;
```
Zylann: A stream can provide both, but this condition will turn off smooth in that case.

terrain/voxel_terrain.cpp
```diff
@@ -894,7 +893,7 @@ void VoxelTerrain::_process() {
 			// Smooth meshing works on more neighbors, so checking a single block isn't enough to ignore it,
 			// but that will slow down meshing a lot.
 			// TODO This is one reason to separate terrain systems between blocky and smooth (other reason is LOD)
-			if (!_smooth_meshing_enabled) {
+			if (_stream->get_used_channels_mask() & 1<<VoxelBuffer::CHANNEL_TYPE) {
```
Zylann: Same here

terrain/voxel_terrain.cpp
```diff
@@ -1060,7 +1059,15 @@ void VoxelTerrain::_process() {
 }
 
 Ref<VoxelTool> VoxelTerrain::get_voxel_tool() {
-	return Ref<VoxelTool>(memnew(VoxelToolTerrain(this, _map)));
+	Ref<VoxelTool> vt = memnew(VoxelToolTerrain(this, _map));
+	if (_stream.is_valid()) {
+		if(_stream->get_used_channels_mask() & 1<<VoxelBuffer::CHANNEL_TYPE) {
```
Zylann: Same here

generators/voxel_generator_test.cpp
tinmanjuggernaut: eh, this was just to get it functional. It currently only supports blocky and doesn't work at all with VT/smooth, or VLT. The whole thing needs work. It doesn't handle LODs on VLT and looks janky. My aim with this PR was not to get it perfect but to get it functional and draw something in every situation.

Zylann: If so, remember to open an issue so it can be tracked

doc/tools/build.py
tinmanjuggernaut: `The slash operator helps create child paths, similarly to os.path.join():`
https://docs.python.org/3/library/pathlib.html#operators

Python3 complained about bindir+filemask and wouldn't concatenate them. I read a recommendation to use /, but then it complained about variables not being a string. Even though yeah they're supposed to be strings, it said one was a path object.

`binfiles = glob.glob(os.path.join(bindir,filemask))` also works and is now being used.

generators/voxel_generator_noise.cpp
tinmanjuggernaut: done

generators/voxel_generator_noise.cpp
tinmanjuggernaut: Yes. And it's only being used here to alert VT so it can set the editor warning flag.

terrain/voxel_lod_terrain.h
tinmanjuggernaut: done

terrain/voxel_terrain.cpp
tinmanjuggernaut: This is just speculation about Godot philosophy in general? I obviously don't know.

I started looking at this with FastNoise. There is the ability to add a recursive lookup. But if you select that option and don't create the sub-object it will crash. So I added the sub-object by default. Similarly, here if you don't add a VoxelLibrary, VT won't display anything. Adding it automatically is one step closer to removing non-display conditions.

doc/08_api-overview.md
tinmanjuggernaut: replaced.

terrain/voxel_terrain.cpp
tinmanjuggernaut: done, with the other two

generators/voxel_generator_test.cpp
tinmanjuggernaut: will do

tinmanjuggernaut:
> > when using a gdscript generator, it won't call the gscript get_used_channels_mask
> That function should be implementable by GDScript, the same way emerge/immerge is.

Got it, thank you.

> > Also note, I've put the ADDPROPERTY channel into the Image/Noise2D child classes, though the variable is in VGheightmap.cpp
> 
> That's quite unfortunate... changing an API just because of editor display problems xD

Is this an issue as it is? I guess it only matters if a GDScript extends VoxelGeneratorHeightmap, they might have to make their own channel variable, or they could just use the setter/getter. If they extend VGImage, they can use the setter/getter or channel=1. If they extend VG, they have to implement their own channel, set/get, and get_used_channels_mask entirely.

I tweaked the channel detector so both VT and VLT flag all unsupported channels.

The PR has been updated. Let me know if there's anything else to adjust.

PR #132, opened 2020-03-28
Fix build warnings on other platforms

tinmanjuggernaut:


.gitignore
```diff
@@ -3,3 +3,4 @@
 *.o
 *.autosave
 *.bc
+*.d
```
tinmanjuggernaut: iphone

meshers/blocky/voxel_library.cpp
```diff
@@ -269,7 +269,7 @@ void VoxelLibrary::generate_side_culling_matrix() {
 
 			CRASH_COND(pattern == nullptr);
 
-			if (full_side_pattern_index == -1 && bitmap.all()) {
+			if (full_side_pattern_index == 0xFFFFFFFF && bitmap.all()) {
```
tinmanjuggernaut: The warning is here. It doesn't compare about assigning -1, only comparing uint vs -1.

SCsub
```diff
@@ -22,6 +22,9 @@ files = [
 for f in files:
 	env_voxel.add_source_files(env.modules_sources, f)
 
+if env['platform'] == 'osx' or env['platform'] == 'android':
+    env_voxel.Append(CXXFLAGS=['-Wno-inconsistent-missing-override'])
```
Zylann: Add comment explaining it's because Godot's codebase is old

meshers/blocky/voxel_library.cpp
```diff
@@ -181,7 +181,7 @@ void VoxelLibrary::generate_side_culling_matrix() {
 	};
 
 	std::vector<Pattern> patterns;
-	uint32_t full_side_pattern_index = -1;
+	uint32_t full_side_pattern_index = 0xFFFFFFFF;
```
Zylann: Could put that `0xffffffff` in a `NULL_INDEX` constant?

tinmanjuggernaut:
Updates committed.

meshers/blocky/voxel_library.h
```diff
@@ -5,6 +5,8 @@
 #include "voxel.h"
 #include <core/resource.h>
 
+#define NULL_INDEX 0xFFFFFFFF
```
Zylann, line 8: Use a `static const uint32_t`, not a macro

tinmanjuggernaut: Updated. Why is this way better?

Zylann:
Thanks!

PR #140, opened 2020-04-24
Added raycast case for integer positions

### LINKED ISSUE #136: VoxelTool Raycast does not work with integer raycast origin
The VoxelTool Raycasting is mean to allow raycasting without needing to use colliders, as this can enhance performance in games that do not require physics.

Unfortunately it does not return the right results. Here I have set up a simple raycast from the mouse:

``` C#
using Godot;
using Godot.Collections;
using System;

public class Spatial : Godot.Spatial
{
    private VoxelTerrain ter;
    private Camera cam;
    private Vector2 mousePos;
    private Vector3? hoveredVoxel;

    public override void _Ready()
    {
        ter = GetNode<VoxelTerrain>("VoxelTerrain");
        cam = GetNode<Camera>("Camera");

        mousePos = Vector2.Zero;
        hoveredVoxel = null;
    }

    public override void _Input(InputEvent @event) {
        if (@event is InputEventMouseMotion mouseMotionEvent) {
            mousePos = mouseMotionEvent.Position;
        }

        if (@event is InputEventMouseButton mouseButton && mouseButton.IsPressed() && (ButtonList)mouseButton.ButtonIndex == ButtonList.Left) {
            GD.Print("Clicked on ", hoveredVoxel);
            ter.GetVoxelTool().SetVoxel(hoveredVoxel.Value, 0);
        }
    }

    public override void _PhysicsProcess(float delta) {
        Vector3 from = cam.ProjectRayOrigin(mousePos);
        Vector3 to = cam.ProjectRayNormal(mousePos);

        VoxelRaycastResult result = ter.GetVoxelTool().Raycast(from, to, 1000);
        if (result != null)
            hoveredVoxel = result.Position;
    }
}
```

[Lets look at the results.](https://streamable.com/hefeex) Excuse the poor visibility — I have yet to get the ambient occlusion working properly as the "simply enable vertex color/as albedo" in the docs appears to be incorrect.

As you can see, the raycast works fine for voxels on the lefthand side of the screen. However, on the right side of the screen, the raycast is off by one or two blocks to the right.

thorlucas:
Closes #136

### DIFF
--- a/util/voxel_raycast.h
+++ b/util/voxel_raycast.h
@@ -75,6 +75,28 @@ bool voxel_raycast(
 	} else
 		tcross_z = g_infinite; // Will never cross on X
 
+	// Workaround for integer positions
+	// Adapted from https://github.com/bulletphysics/bullet3/blob/3dbe5426bf7387e532c17df9a1c5e5a4972c298a/src/BulletCollision/CollisionShapes/btHeightfieldTerrainShape.cpp#L418
+	if (tcross_x == 0.0) {
+		tcross_x += tdelta_x;
+		// If going backwards, we should ignore the position we would get by the above flooring,
+		// because the ray is not heading in that direction
+		if (xi_step == -1)
+			hit_pos.x -= 1;
+	}
+
+	if (tcross_y == 0.0) {
+		tcross_y += tdelta_y;
+		if (yi_step == -1)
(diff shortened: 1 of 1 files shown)


Zylann:
Thanks!

PR #141, opened 2020-04-29
Generate a collision mesh for each surface

nwdamgaard:
I was trying to make a game and it was driving me mad how material 1 is the only one that gets a collision mesh, so I fixed it. Unfortunately, somehow I added a bug where the mesh for the collision debug view doesn't update anymore, and it's a little out of my wheelhouse to fix. I'm sure that the issue is that I'm not using Ref<>'s correctly in direct_static_body.cpp while adding multiple shape support but I don't know much about Ref<>'s and I'm quite lazy so there's that.

### DIFF
--- a/meshers/blocky/voxel_mesher_blocky.cpp
+++ b/meshers/blocky/voxel_mesher_blocky.cpp
@@ -48,6 +48,285 @@ inline bool contributes_to_ao(const VoxelLibrary &lib, int voxel_id) {
 } // namespace
 
 template <typename Type_T>
+static void generate_collision_surface(
+		VoxelMesherBlocky::Arrays &arrays,
+		const ArraySlice<Type_T> type_buffer,
+		const Vector3i block_size,
+		const VoxelLibrary &library,
+		bool bake_occlusion, float baked_occlusion_darkness) {
+
+	// Build lookup tables so to speed up voxel access.
+	// These are values to add to an address in order to get given neighbor.
+
+	int row_size = block_size.y;
+	int deck_size = block_size.x * row_size;
+
+	// Data must be padded, hence the off-by-one
+	Vector3i min = Vector3i(VoxelMesherBlocky::PADDING);
+	Vector3i max = block_size - Vector3i(VoxelMesherBlocky::PADDING);
+
+	int index_offsets[VoxelMesherBlocky::MAX_MATERIALS] = { 0 };
+
+	FixedArray<int, Cube::SIDE_COUNT> side_neighbor_lut;
+	side_neighbor_lut[Cube::SIDE_LEFT] = row_size;
+	side_neighbor_lut[Cube::SIDE_RIGHT] = -row_size;
+	side_neighbor_lut[Cube::SIDE_BACK] = -deck_size;
+	side_neighbor_lut[Cube::SIDE_FRONT] = deck_size;
+	side_neighbor_lut[Cube::SIDE_BOTTOM] = -1;
+	side_neighbor_lut[Cube::SIDE_TOP] = 1;
+
+	FixedArray<int, Cube::EDGE_COUNT> edge_neighbor_lut;
+	edge_neighbor_lut[Cube::EDGE_BOTTOM_BACK] = side_neighbor_lut[Cube::SIDE_BOTTOM] + side_neighbor_lut[Cube::SIDE_BACK];
+	edge_neighbor_lut[Cube::EDGE_BOTTOM_FRONT] = side_neighbor_lut[Cube::SIDE_BOTTOM] + side_neighbor_lut[Cube::SIDE_FRONT];
+	edge_neighbor_lut[Cube::EDGE_BOTTOM_LEFT] = side_neighbor_lut[Cube::SIDE_BOTTOM] + side_neighbor_lut[Cube::SIDE_LEFT];
+	edge_neighbor_lut[Cube::EDGE_BOTTOM_RIGHT] = side_neighbor_lut[Cube::SIDE_BOTTOM] + side_neighbor_lut[Cube::SIDE_RIGHT];
+	edge_neighbor_lut[Cube::EDGE_BACK_LEFT] = side_neighbor_lut[Cube::SIDE_BACK] + side_neighbor_lut[Cube::SIDE_LEFT];
+	edge_neighbor_lut[Cube::EDGE_BACK_RIGHT] = side_neighbor_lut[Cube::SIDE_BACK] + side_neighbor_lut[Cube::SIDE_RIGHT];
+	edge_neighbor_lut[Cube::EDGE_FRONT_LEFT] = side_neighbor_lut[Cube::SIDE_FRONT] + side_neighbor_lut[Cube::SIDE_LEFT];
+	edge_neighbor_lut[Cube::EDGE_FRONT_RIGHT] = side_neighbor_lut[Cube::SIDE_FRONT] + side_neighbor_lut[Cube::SIDE_RIGHT];
+	edge_neighbor_lut[Cube::EDGE_TOP_BACK] = side_neighbor_lut[Cube::SIDE_TOP] + side_neighbor_lut[Cube::SIDE_BACK];
+	edge_neighbor_lut[Cube::EDGE_TOP_FRONT] = side_neighbor_lut[Cube::SIDE_TOP] + side_neighbor_lut[Cube::SIDE_FRONT];
+	edge_neighbor_lut[Cube::EDGE_TOP_LEFT] = side_neighbor_lut[Cube::SIDE_TOP] + side_neighbor_lut[Cube::SIDE_LEFT];
+	edge_neighbor_lut[Cube::EDGE_TOP_RIGHT] = side_neighbor_lut[Cube::SIDE_TOP] + side_neighbor_lut[Cube::SIDE_RIGHT];
+
+	FixedArray<int, Cube::CORNER_COUNT> corner_neighbor_lut;
+
+	corner_neighbor_lut[Cube::CORNER_BOTTOM_BACK_LEFT] =
+			side_neighbor_lut[Cube::SIDE_BOTTOM] +
+			side_neighbor_lut[Cube::SIDE_BACK] +
+			side_neighbor_lut[Cube::SIDE_LEFT];
+
+	corner_neighbor_lut[Cube::CORNER_BOTTOM_BACK_RIGHT] =
+			side_neighbor_lut[Cube::SIDE_BOTTOM] +
+			side_neighbor_lut[Cube::SIDE_BACK] +
+			side_neighbor_lut[Cube::SIDE_RIGHT];
+
+	corner_neighbor_lut[Cube::CORNER_BOTTOM_FRONT_RIGHT] =
+			side_neighbor_lut[Cube::SIDE_BOTTOM] +
+			side_neighbor_lut[Cube::SIDE_FRONT] +
+			side_neighbor_lut[Cube::SIDE_RIGHT];
+
+	corner_neighbor_lut[Cube::CORNER_BOTTOM_FRONT_LEFT] =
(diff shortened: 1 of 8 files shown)


nwdamgaard:
As an additional note, it seems that once you start using more than one collision surface and you put a rigidbody on top, it tanks performance. A better approach to solving this issue would be to add a function to the mesher to generate collision meshes.

Zylann:
> As an additional note, it seems that once you start using more than one collision surface and you put a rigidbody on top, it tanks performance.

Assuming you are using the blocky mesher (which is the only one supporting multiple surfaces), surfaces are usually separate. If you want them to be treated as a single one, they have to be batched as one, not many. Not sure why not doing that tanks performance, though.

Also note Minecraft-like worlds usually don't require classic collision shapes since AABB physics can be used, which is immensely simpler and faster. Realistic physics over a large blocky world can be quite costy :p

> A better approach to solving this issue would be to add a function to the mesher to generate collision meshes.0

Which will tank performance too^^" As I said the fastest is to take the surfaces that were generated already and batch them.

Can you rebase your PR, perhaps? It shows you added the `VoxelBlock::set_mesh` function entirely, but I doubt you wrote it originally so I'm not sure which parts you actually changed

meshers/blocky/voxel_mesher_blocky.cpp
```diff
@@ -47,6 +47,285 @@ inline bool contributes_to_ao(const VoxelLibrary &lib, int voxel_id) {
...
+						}
+
+						index_offset += vertex_count;
+					}
+				}
+			}
+		}
+	}
+}
+
```
Zylann, line 328: Please, don't do this. Don't copy/paste this entire function. Besides, you are forcing blocky mesher for every kind of terrain. I wrote a comment earlier explaining how it should be done. The meshes that were generated *already* need to be used and batched, not recomputed.

nwdamgaard:
>Also note Minecraft-like worlds usually don't require classic collision shapes since AABB physics can be used, which is immensely simpler and faster. Realistic physics over a large blocky world can be quite costy :p

You see, I'm trying to make a demo/game with smaller voxels, so to achieve that I made the character bigger to make the voxels appear smaller. Now, it would be quite annoying to have to jump up every little step, so I'm using a capsule collider that way it slides up small inclines.

> > A better approach to solving this issue would be to add a function to the mesher to generate collision meshes.

>Which will tank performance too^^" As I said the fastest is to take the surfaces that were generated already and batch them.

Alright so before I saw this message I spent all afternoon adding that mesher function I was talking about. It does speed up the rigidbody collisions, but, as you said, it slows down mesh generation.

>Don't copy/paste this entire function.

How do you suggest I do it?

Zylann:
> You see, I'm trying to make a demo/game with smaller voxels, so to achieve that I made the character bigger to make the voxels appear smaller. Now, it would be quite annoying to have to jump up every little step, so I'm using a capsule collider that way it slides up small inclines.

Fair enough. Do you still need the collisions to be blocky? That sounds quite misfit for what you want  to do. Although, this module does not support entirely different meshes from the visual ones (due to performance cost of generating said meshes, mostly).

> Alright so before I saw this message I spent all afternoon adding that mesher function I was talking about. It does speed up the rigidbody collisions, but, as you said, it slows down mesh generation.

Is it in your PR? What does it do compared to existing meshing functions?

> How do you suggest I do it?

`set_mesh` receives surface arrays inside its `surface_arrays` parameter. The problem, as you said, is that it corresponds to only the first surface. But I believe no meshing should happen inside, because:

1) It would be on the main thread, so it slows the game down
2) You'd have to inject the actual mesher used, because `VoxelBlock` is used for all kinds of terrains, not just blocky ones
3) It would polygonize voxels a second time, which is essentially redundant, because it was done already. Besides, polygonizing can require neighbor voxels, which is extra boilerplate to setup in the wrong place.

The solution is to provide all surfaces to the function, eventually with some metadata to decide which to pick, which layer to use etc.
In the code calling `set_mesh` (in `VoxelTerrain` and `VoxelLodTerrain`) there is a point where we turn a `Vector<Array>` containing polygonized surfaces into a `Ref<Mesh>`, while also keeping a reference to one `Array` only, considering it the collidable surface. Instead, it should keep the whole `Vector<Array>`, and send that to `set_mesh`. 
Then in `set_mesh`, there is a `create_concave_polygon_shape` function which turns arrays into `Ref<Shape>`. This function should be changed such that it can take multiple surfaces, and treat them as one, so that it still returns a single shape, containing polygons of all surfaces sent to it. Eventually, you may want to ignore some of the surfaces (if they are water, or grass for example), or even choose which go into which shape, if multiple collision layers are needed. That's the role of metadata. However that means there is a bit more changes and design to make so that collision can be better configured.

nwdamgaard:
Okay, I present to you, the final solution. Here's how it works:

- I added `VoxelMesherBlocky::generate_collision_surface()`. What it does is it takes each mesh per material and combines them into just one mesh.
- This is called from `VoxelMesherBlocky::build()`, which as you know runs on the mesh updater thread.
- The mesh is delivered back to `VoxelTerrain` via a field I added to `VoxelMesher::Output`
- The mesh is made into the collision shape by passing it into `VoxelBlock::set_mesh` in a parameter that was already used to set the collision shape so I don't know what happens to it beyond there.

Zylann:
> I added VoxelMesherBlocky::generate_collision_surface(). What it does is it takes each mesh per material and combines them into just one mesh.

You should not need to modify `VoxelMesherBlocky`... it polygonizes voxels into surfaces for sure, but that's not where collision handling should happen IMO. The function to modify is at later stage, `create_concave_polygon_shape`. Besides, the code you added allocates double memory just to copy the data a second time, including arrays that are not even required for collision (uv? color?). Sure, vertices get put together, but it should not involve an entire copy. That data already exists, is already made. All that needs to change is the way it gets used. Also following this route there are technically 2 other meshers that would have to implement this, it's not worth it.
Really, just use the same arrays that are used to build the `Ref<Mesh>`, they already contain the exact same data you copied. Using a loop in `create_concave_polygon_shape` instead of the single array lookup should do the trick and avoid the copies.

terrain/voxel_block.cpp
```diff
@@ -7,27 +7,32 @@
 // Faster version of Mesh::create_trimesh_shape()
 // See https://github.com/Zylann/godot_voxel/issues/54
 //
-static Ref<ConcavePolygonShape> create_concave_polygon_shape(Array surface_arrays) {
+static Ref<ConcavePolygonShape> create_concave_polygon_shape(Vector<Array*> surfaces) {
```
Zylann, line 10: No need for it to be `Array*`, the Godot `Array` type is reference-counted so it can be passed around for very little cost. Besides, Godot's `Vector<T>` also works that way so that's why it is also passed without `*` or `&`.

terrain/voxel_block.cpp
```diff
@@ -7,27 +7,32 @@
...
+		ERR_FAIL_COND_V(indices.size() < 3, Ref<ConcavePolygonShape>());
+		ERR_FAIL_COND_V(indices.size() % 3 != 0, Ref<ConcavePolygonShape>());
 
-	{
-		PoolVector<Vector3>::Write w = face_points.write();
-		PoolVector<int>::Read index_r = indices.read();
-		PoolVector<Vector3>::Read position_r = positions.read();
+		int face_points_count = indices.size();
+		int face_points_start = face_points.size();
+		face_points.resize(face_points_start + face_points_count);
```
Zylann, line 26: Could resize only once by summing surface sizes after a first pass

terrain/voxel_block.cpp
```diff
@@ -7,27 +7,32 @@
...
 
-		for (int i = 0; i < face_points_count; ++i) {
-			w[i] = position_r[index_r[i]];
+		{
+			PoolVector<Vector3>::Write w = face_points.write();
+			PoolVector<int>::Read index_r = indices.read();
+			PoolVector<Vector3>::Read position_r = positions.read();
+
+			for (int i = face_points_start; i < face_points_count; ++i) {
+				w[i] = position_r[index_r[i]];
```
Zylann, line 34: When this code gets to the second surface, `i` will no longer start from the beginning of the surface, resulting in buffer overrun. You should probably use a second counter variable.

terrain/voxel_lod_terrain.cpp
```diff
@@ -14,7 +14,7 @@ const uint32_t MAIN_THREAD_MESHING_BUDGET_MS = 8;
 namespace {
 
 Ref<ArrayMesh> build_mesh(const Vector<Array> surfaces, Mesh::PrimitiveType primitive, int compression_flags,
-		Ref<Material> material, Array *collidable_surface) {
+		Ref<Material> material, Vector<Array*> &collidable_surface) {
```
Zylann, line 17: I believe this can be `Vector<Array>*`

terrain/voxel_lod_terrain.cpp
```diff
@@ -32,8 +32,8 @@ Ref<ArrayMesh> build_mesh(const Vector<Array> surfaces, Mesh::PrimitiveType prim
 			continue;
 		}
 
-		if (collidable_surface != nullptr && collidable_surface->empty()) {
-			*collidable_surface = surface;
+		if (collidable_surface[0]->empty()) {
+			*collidable_surface[0] = surface;
```
Zylann, line 36: And then this can be `collidable_surfaces != nullptr`. tbh I might do a second pass on this later because collision support may need some refinements beyond support for multiple surfaces

terrain/voxel_lod_terrain.cpp
```diff
@@ -1198,12 +1198,12 @@ void VoxelLodTerrain::_process() {
 			const VoxelMesher::Output mesh_data = ob.data.smooth_surfaces;
 
 			// TODO Allow multiple collision surfaces
-			Array collidable_surface;
+			Vector<Array*> collidable_surface;
```
Zylann, line 1201: In fact there is probably no need to create an extra vector, you could use `mesh_data.surfaces` directly? Then no need for the last argument either

terrain/voxel_terrain.cpp
```diff
@@ -1018,15 +1018,13 @@ void VoxelTerrain::_process() {
 					continue;
 				}
 
+				collidable_surfaces.push_back((Array*)&data.blocky_surfaces.surfaces[i]);
```
Zylann, line 1021: No need to cast because it should not need to be a pointer type, as said earlier

tinmanjuggernaut:
> Now, it would be quite annoying to have to jump up every little step, so I'm using a capsule collider that way it slides up small inclines.

You can also use a raycast collision shape to smooth it out. This is generally recommended in the Godot docs and is what my demo uses.

If you only need physics for the character, you can replace it entirely with a grid of raycasts around the character, or a combo of raycasts and AABB.

nwdamgaard:
>replace it entirely with a grid of raycasts around the character, or a combo of raycasts and AABB.

That sounds like a very sensible solution, but now that I've spent hours patching this in I want to do it this way

terrain/voxel_terrain.cpp
```diff
@@ -1018,15 +1018,13 @@ void VoxelTerrain::_process() {
 					continue;
 				}
 
+				collidable_surfaces.push_back(data.blocky_surfaces.surfaces[i]);
```
Zylann, line 1021: Just use `surface`. Also, prefer keeping this after the `is_surface_triangulated(surface)`.

terrain/voxel_terrain.cpp
```diff
@@ -1039,15 +1037,13 @@ void VoxelTerrain::_process() {
 					continue;
 				}
 
+				collidable_surfaces.push_back(data.smooth_surfaces.surfaces[i]);
```
Zylann, line 1040: Just use `surface`. Also, prefer keeping this after the `is_surface_triangulated(surface)`.

PR #152, opened 2020-07-06
Fixed cannot place blocks on transparent blocks

nic96:
Not sure if this is the right way of fixing this, but I have some transparent blocks like glass or leaves that I still want to place other blocks onto, but the raycast method ignored them. This is my quick, but there might be better solution like maybe having some sort of raycast option in the VoxelLibrary.

Closes issue #153

### DIFF
--- a/edition/voxel_tool_terrain.cpp
+++ b/edition/voxel_tool_terrain.cpp
@@ -40,6 +40,9 @@ Ref<VoxelRaycastResult> VoxelToolTerrain::raycast(Vector3 pos, Vector3 dir, floa
 			if (voxel.is_transparent() == false)
 				return true;
 
+			if (voxel.is_transparent() && voxel.get_collision_aabbs().empty() == false)
+				return true;
+
 			float v1 = map->get_voxel_f(pos, VoxelBuffer::CHANNEL_SDF);
 			return v1 < 0;
 		}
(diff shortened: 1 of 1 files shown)


edition/voxel_tool_terrain.cpp
```diff
@@ -37,9 +37,15 @@ Ref<VoxelRaycastResult> VoxelToolTerrain::raycast(Vector3 pos, Vector3 dir, floa
 				return false;
 
 			const Voxel &voxel = lib.get_voxel_const(v0);
+			if (voxel.get_geometry_type() == voxel.GEOMETRY_NONE)
+				return false;
+
 			if (voxel.is_transparent() == false)
 				return true;
 
+			if (voxel.is_transparent() && voxel.get_collision_aabbs().empty() == false)
```
Zylann, line 46: Technically, if the voxel has collision AABBs, an optional narrow phase should be run on them with a for loop I believe, because as in minecraft, a voxel model is not necessarily a cube. Yet water has a collision AABB, just not taken into account so there will still be room for refinements in raycast capabilities.
Your code should be enough for now, but it's going to change in the future anyways.

nic96:
So maybe raycast method should have an option to exclude block IDs or include only given IDs.

Zylann:
By ID would probably be too much. Eventually a version taking a funcref to do the hit test would allow full customization, but for now the collision AABBs should be enough. Maybe voxel types can be tagged (similarly to collision masks) so they can get ignored.

PR #154, opened 2020-07-11
Fix includes

nic96:


Zylann:
I already pushed `<limits>` to Interval and removed usages of the C limits to use the C++ ones instead.

nic96:
Perfect thanks.

Issue #146, opened 2020-05-20
Ability to save changes without using LOD?


wacyym:
Can i save terrain changes using VoxelTerrain (not VoxelTerrainLod)?
I try to do this with this settings using VoxelStreamRegionFiles:
![image](https://user-images.githubusercontent.com/6958185/82463834-5ea47c80-9ace-11ea-87ee-8ccd9c6f6c98.png)

But changes not saved after restart level.

Edit: using VoxelTerrainLOD + VoxelStreamRegionFiles works...

Zylann:
Do you get any errors?

menip:
Are we supposed to be able to save/load blocky terrain (`VoxelTerrain`) with, for example, `VoxelStreamRegionFiles`? 
I've tried, but it seems to not work. As shown in OPs screenshot, `VoxelTerrain` has warning saying "`VoxelTerrain` supports only stream channels 'Type' or 'SDF'."

There are no errors being printed to console. What I noticed is that when option `Save Fallback Output` is enabled, some files do get saved. When it is not, nothing gets saved. Additionally, terrain modifications don't get saved: try to add/remove voxels, restart level, then you see that only terrain from fallback stream is there, modifications are not.


I was originally thinking I have to make my own generator to have persistent terrain modifications, but would much prefer for a built in solution, such as `VoxelStreamRegionFiles`.

Zylann:
> Are we supposed to be able to save/load blocky terrain (VoxelTerrain) with, for example, VoxelStreamRegionFiles?

Yes.

> What I noticed is that when option Save Fallback Output is enabled, some files do get saved

That option saves blocks as they are generated. When turned off, it only saves modified blocks and generates the others again in the next session.

If you run the game in verbose mode (`-v`) you should see a console log when a block is detected as being modified. They are normally saved when they get unloaded. If not then there is probably a bug in `VoxelTerrain`.

menip:
>They are normally saved when they get unloaded. I'm not sure if they will if you use the Stop button from the editor (as it kills the game ungracefully?).

Ah, this was the important bit. I tried modifying blocks, flying out of view distance so my modified block gets unloaded, then restarting game. Modifications were loaded as expected. However, changes do not get saved by just exiting the game, gracefully or not. I expected that by default, if you exit any any way, changes would be saved (autosaving, but I guess this can just be triggered via script).

Zylann:
[...]

PR #161, opened 2020-07-24
Fix autosave on window close not working

menip:
See https://github.com/Zylann/godot_voxel/issues/146, this PR fixes the autosaving bug. What was happening, was the truth value of `data.thread_exit` could be changed by main thread while child thread was inside loop.

### DIFF
--- a/terrain/block_thread_manager.h
+++ b/terrain/block_thread_manager.h
@@ -374,7 +374,7 @@ private:
 
 	static void thread_func(JobData &data) {
 
-		while (!data.thread_exit) {
+		while (true) {
 
 			uint32_t sync_time = OS::get_singleton()->get_ticks_msec() + data.sync_interval_ms;
 


terrain/block_thread_manager.h
```diff
@@ -466,12 +467,14 @@ class VoxelBlockThreadManager {
...
 
-			if (data.thread_exit) {
+			if (data.thread_exit && data.semaphore->get() >= 1) {
 				break;
 			}
 
 			// Wait for future wake-up
 			data.semaphore->wait();
+		
+			keep_going = !data.thread_exit || data.semaphore->get() > 0;
```
Zylann, line 477: I'm not sure of this. The reason why the semaphore is raised is to make the call to `wait()` resume the thread, so it will always be true here I think?
I believe the exit condition doesn't need to be driven by the semaphore.

menip: Lets assume normally when we enter thread, semaphore has value of 1. All work is done, and we reach line 475. After this line runs, semaphore will have value of 0, right?

However, I noticed the above is not always the case, frequently when autosaving, for some reason semaphore value is 2, and every time this happens, I noticed that my changes weren't saved.  Thus, the reason for `keep_going = !data.thread_exit || data.semaphore->get() > 0;`.

What honestly confuses me more, is that from further testing, it seems the following change is not needed at all: `if (data.thread_exit && data.semaphore->get() >= 1)`.

I'll have the updated PR shortly.

Zylann: I wonder if it doesnt just need `while(true)` on the outer loop instead of `while(!thread_exit)`, with a `break` if `thread_exit == true` in the place you added it.

menip: It seems you are correct :laughing: 
The only change needed is `while(!data.thread_exit) -> while(true)`.
I'll have that update in just a bit. 

Can you say why this is likely the case? I mean, how can the value of tread_exit change inside the loop?

Zylann: It can change because the main thread can set it anytime. It might be false at the beginning and become true while the loop is processing. That means while that processing happens, some new inputs could have arrived as well, so at the end the loop needs to restart one last time to be sure to catch what remains.

menip: Oh I see. Then I should move the break statement to be last in the loop, even after the semaphore wait call, right?

menip:
I just saw in https://github.com/Zylann/godot_voxel/issues/146#issuecomment-663443448 that Zylann is planning on fixing the code properly, so if we want we can close this PR in favor of proper fix.

terrain/block_thread_manager.h
Zylann: Yeah, cuz it doesn't give the last shot as I described earlier.

Zylann:
I think 638f43f93a307afaf12676bb8ecbf208fc7d773a fixed it

menip:
@Zylann from my testing, eef3818 does indeed seem to fix the bug. I just have one question from the following:

```
bool should_exit = false;
if (data.thread_exit) {
	// This will be the last run, we must process all mandatory requests.
	// We can't check `thread_exit` afterwards because it could have become `true` after we sync,
	// hence we could miss some of the requests.
	should_exit = true;
}
```

The comment says we can't check `thread_exit` after syncing, because it may of become `true`. Did you mean to say it may of become `false`? As in, at the place of the comment, `thread_exit` is already `true`, so after sync it can't change state to  `true`, it already is `true`.

Zylann:
What I mean in the comment is what would happen if we did *not* have this check before the inner loop.
Otherwise yes, when it becomes `true` it should remain. Becoming `false` would be invalid anyways.

PR #169, opened 2020-08-02
fix compilation errors with mingw compiler + show terrain in editor +…

bhohbaum:
… prevent editor crash when streaming directory path is empty

there are still issues with the culling, the settings seem to be ignored in the editor, but it is usable.

### DIFF
--- a/streams/voxel_stream_region_files.cpp
+++ b/streams/voxel_stream_region_files.cpp
@@ -508,6 +508,7 @@ static void migrate_region_meta_data(Dictionary &data) {
 VoxelFileResult VoxelStreamRegionFiles::load_meta() {
 
 	ERR_FAIL_COND_V(_directory_path == "", VOXEL_FILE_CANT_OPEN);
+	if (_directory_path.empty()) return VOXEL_FILE_CANT_OPEN;
 
 	// Ensure you cleanup previous world before loading another
 	CRASH_COND(_region_cache.size() > 0);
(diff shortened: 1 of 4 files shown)


Zylann:
Don't include the changes to show terrain in editor. This is a task I have to do properly at some point, there are a number of dangerous situations that can arise if turned on at the moment.

bhohbaum:
That is totally possible. I just catched one issue (the possibly empty dir
param), that happened to pop up on my test project. But this definitely
needs more testing and I'm not happy with some other points as well like my
coding guidelines coherence.... I didn't start this with the intention to
publish my code.

I wasn't sure why this feature is not implemented yet. So I thought I
implement the core of it and someone with better knowledge of the module
does the fine tuning and possibly additional features to optimize
usability/overall concept/stability.

I will probably play around with this some more and will develop it
further. In case you're interested in this, I'll push my updates to this
feature in the branch associated with this pull request. We will see if it
reaches a point where it can be merged or used as a base for further
development.

Just let me know.




Marc <notifications@github.com> schrieb am So., 2. Aug. 2020, 19:51:

streams/voxel_stream_region_files.cpp
```diff
@@ -508,6 +508,7 @@ static void migrate_region_meta_data(Dictionary &data) {
 VoxelFileResult VoxelStreamRegionFiles::load_meta() {
 
 	ERR_FAIL_COND_V(_directory_path == "", VOXEL_FILE_CANT_OPEN);
+	if (_directory_path.empty()) return VOXEL_FILE_CANT_OPEN;
```
menip, line 511: I'm pretty sure this if statement is never going to run, as `ERR_FAIL_COND_V` already returns in the case that path is empty.

Zylann:
Closing as I think this was adressed

PR #193, opened 2020-09-07
Fix binding of is_empty to allow build with mono

tdaffin:
In order to build godot with godot_voxel and mono I had to make the change to the is_empty binding in meshers/blocky/voxel.cpp
Without this change the cs file generated during `./bin/godot.x11.tools.64.mono --generate-mono-glue modules/mono/glue` is not valid C# code.
I'm not sure if this will cause or fix a similar problem creating other bindings -- for gdscript for instance.
This is my first contribution to anything godot related, so this change is more of an educated guess than anything :)

### DIFF
--- a/meshers/blocky/voxel.cpp
+++ b/meshers/blocky/voxel.cpp
@@ -418,7 +418,7 @@ void Voxel::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("set_collision_mask", "mask"), &Voxel::set_collision_mask);
 	ClassDB::bind_method(D_METHOD("get_collision_mask"), &Voxel::get_collision_mask);
 
-	ClassDB::bind_method(D_METHOD("is_empty()"), &Voxel::is_empty);
+	ClassDB::bind_method(D_METHOD("is_empty"), &Voxel::is_empty);
 
 	// TODO Update to StringName in Godot 4
 	ADD_PROPERTY(PropertyInfo(Variant::STRING, "voxel_name"), "set_voxel_name", "get_voxel_name");


tdaffin:
I created a test project with the build I made with this change and was able to use Voxel.is_empty in gdscript and Voxel.IsEmpty() in C#, so I guess that this change is ok.
Project is very simple with just a VoxelTerrain and Camera as children of a Spatial with the scripts on the Spatial.

GDScript:
`
extends Spatial

onready var terrain = $VoxelTerrain

func _ready():
	var voxelLibrary = terrain.voxel_library
	
	print("Testing is_empty in gdscript")

	print("NumVoxels=" + str(voxelLibrary.voxel_count))
	for id in range(voxelLibrary.voxel_count):
		var voxel = voxelLibrary.get_voxel(id)
		print("Voxel " + str(id) + ": " + voxel.voxel_name + " " + str(voxel.is_empty()))
`

C#:
`
using Godot;
using System;

public class Spatial : Godot.Spatial
{
    // Called when the node enters the scene tree for the first time.
    public override void _Ready()
    {
        var terrain = GetNode<VoxelTerrain>("VoxelTerrain");
        var voxelLibrary = terrain.VoxelLibrary;

        GD.Print("Testing is_empty in C#");

        GD.Print($"NumVoxels={voxelLibrary.VoxelCount}");
        for (uint id = 0, n = voxelLibrary.VoxelCount; id < n; ++id)
        {
            var voxel = voxelLibrary.GetVoxel(id);
            GD.Print($"Voxel {id}: {voxel.VoxelName} {voxel.IsEmpty()}");
        }
    }
}
`

Zylann:
For sure that was a typo.
Thanks!

PR #195, opened 2020-09-19
Fix virtual bindings to work with C#

tdaffin:
This makes it possible to write a custom voxel generator in C#.
I've posted an example to test with here: https://github.com/tdaffin/voxel_tools_mono

You'll have to build Godot with a patched mono runtime for this to work, because the C++ code is making a call into C# from a thread created outside of C#.  Without the patched runtime or you'll run a crash in unpatched mono. 
 For details see this PR: https://github.com/godotengine/godot/pull/42039

### DIFF
--- a/generators/voxel_generator.cpp
+++ b/generators/voxel_generator.cpp
@@ -31,6 +31,8 @@ void VoxelGenerator::_b_generate_block(Ref<VoxelBuffer> out_buffer, Vector3 orig
 
 void VoxelGenerator::_bind_methods() {
 	// Note: C++ inheriting classes don't need to re-bind these, because they are bindings that call the actual virtual methods
-
-	ClassDB::bind_method(D_METHOD("generate_block", "out_buffer", "origin_in_voxels", "lod"), &VoxelGenerator::_b_generate_block);
+	BIND_VMETHOD(MethodInfo("generate_block",
+							PropertyInfo(Variant::OBJECT, "out_buffer", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_DEFAULT, "VoxelBuffer"),
+							PropertyInfo(Variant::VECTOR3, "origin_in_voxels"),
+							PropertyInfo(Variant::INT, "lod")));
 }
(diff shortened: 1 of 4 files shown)


generators/voxel_generator.cpp
```diff
@@ -31,6 +31,8 @@ void VoxelGenerator::_b_generate_block(Ref<VoxelBuffer> out_buffer, Vector3 orig
 
 void VoxelGenerator::_bind_methods() {
 	// Note: C++ inheriting classes don't need to re-bind these, because they are bindings that call the actual virtual methods
-
-	ClassDB::bind_method(D_METHOD("generate_block", "out_buffer", "origin_in_voxels", "lod"), &VoxelGenerator::_b_generate_block);
+	BIND_VMETHOD(MethodInfo("generate_block",
```
Zylann, line 34: Looks like this will make it impossible to call `generate_block` anymore from script?

streams/voxel_stream.cpp
```diff
@@ -70,9 +70,14 @@ bool VoxelStream::has_script() const {
 }
 
 void VoxelStream::_bind_methods() {
-	// TODO Make these proper virtual, it confuses C# bindings
 	// Note: C++ inheriting classes don't need to re-bind these, because they are bindings that call the actual virtual methods
-	ClassDB::bind_method(D_METHOD("emerge_block", "out_buffer", "origin_in_voxels", "lod"), &VoxelStream::_emerge_block);
-	ClassDB::bind_method(D_METHOD("immerge_block", "buffer", "origin_in_voxels", "lod"), &VoxelStream::_immerge_block);
-	ClassDB::bind_method(D_METHOD("get_used_channels_mask"), &VoxelStream::_get_used_channels_mask);
+	BIND_VMETHOD(MethodInfo("emerge_block",
```
Zylann, line 74: Same here, if no method pointers are specified, none of the voxel streams will be usable from scripts unless you inherit them with a script, which is not always the case

Zylann:
I changed the API to properly bind virtual script methods in 10d6e5bd8ab1ba0bddd8640edc3c1e73a2ad3c08
Let me know if it fixes the problem for you

tdaffin:
Thanks for taking care of that.  I'd come up with a version that had both the callable methods and overridables using _ which was only working under some specific circumstances so it wan't really ready for submission.
I'm working with your new changes now, but running into a few problems.
Anyhow, I'll close this PR and start anew on the latest code.

PR #207, opened 2020-11-07
Fix bindings so that generate-mono-glue for C# bindings works

tdaffin:
I had to make a few changes to the latest version so that I could run
`./bin/godot.x11.opt.tools.64.mono --generate-mono-glue modules/mono/glue`
to generate the C# bindings.

Without these changes the command exits with:
```
Generating Voxel.cs...
ERROR: _generate_cs_property: Condition "getter->return_type.cname != setter->arguments.back()->get().type.cname" is true. Returned: ERR_BUG
   At: modules/mono/editor/bindings_generator.cpp:1428.
ERROR: _generate_cs_type: Failed to generate property 'transparency_index' for class 'Voxel'.
   At: modules/mono/editor/bindings_generator.cpp:1298.
ERROR: generate_cs_api: Generation of the Core API C# project failed.
   At: modules/mono/editor/bindings_generator.cpp:1114.
ERROR: handle_cmdline_args: --generate-mono-glue: Failed to generate the C# API.
   At: modules/mono/editor/bindings_generator.cpp:3182.
```

### DIFF
--- a/meshers/blocky/voxel.cpp
+++ b/meshers/blocky/voxel.cpp
@@ -410,7 +410,7 @@ void Voxel::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("set_transparent", "transparent"), &Voxel::set_transparent);
 	ClassDB::bind_method(D_METHOD("is_transparent"), &Voxel::is_transparent);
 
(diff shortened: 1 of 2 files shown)


Zylann:
Thanks!

PR #209, opened 2020-11-09
Added normal map support to blocky & cube meshers by implementing tangent calculations & passing mesh tangents.

wrathofrathma:


meshers/blocky/voxel.cpp
```diff
@@ -224,12 +230,52 @@ static void bake_cube_geometry(Voxel &config, Voxel::BakedData &baked_data, int
...
+	std::vector<Vector2> deltaUV;
+
 	for (unsigned int side = 0; side < Cube::SIDE_COUNT; ++side) {
 		baked_data.model.side_uvs[side].resize(4);
 		std::vector<Vector2> &uvs = baked_data.model.side_uvs[side];
 		for (unsigned int i = 0; i < 4; ++i) {
 			uvs[i] = (config.get_cube_tile(side) + uv[i]) * s;
 		}
+
+		//Tangent calculation modeled after http://www.opengl-tutorial.org/intermediate-tutorials/tutorial-13-normal-mapping/#computing-the-tangents-and-bitangents
```
Zylann, line 242: I'm not very familiar with this, but it looks very complicated for a simple cube. The code here uses lookup tables with precalculated values, and UVs are deterministic as well, so why not just do that?

meshers/blocky/voxel.cpp
```diff
@@ -224,12 +230,52 @@ static void bake_cube_geometry(Voxel &config, Voxel::BakedData &baked_data, int
 	CRASH_COND(atlas_size <= 0);
 	const float s = 1.0 / atlas_size;
 
+	std::vector<Vector2> deltaUV;
```
Zylann, line 233: Naming convention should be snake_case

meshers/blocky/voxel.cpp
```diff
@@ -190,6 +190,8 @@ void Voxel::set_collision_mask(uint32_t mask) {
 static void bake_cube_geometry(Voxel &config, Voxel::BakedData &baked_data, int p_atlas_size) {
 	const float sy = 1.0;
 
+	std::vector<Vector3> deltaPos;
```
Zylann, line 193: This function is specialized for cubes. You don't need a dynamic vector. You could use a `FixedArray<Vector3, Cube::SIDE_COUNT * 3>` instead.
But even then, I'm wonder why you need to hold data for every side at once. Each face is an independent quad, so I'm sure you could use an array of 3 elements, or a lookup table of precalculated tangents.

meshers/blocky/voxel.cpp
```diff
@@ -253,6 +299,7 @@ static void bake_mesh_geometry(Voxel &config, Voxel::BakedData &baked_data) {
 	PoolVector3Array positions = arrays[Mesh::ARRAY_VERTEX];
 	PoolVector3Array normals = arrays[Mesh::ARRAY_NORMAL];
 	PoolVector2Array uvs = arrays[Mesh::ARRAY_TEX_UV];
+	PoolVector<real_t> tangents = arrays[Mesh::ARRAY_TANGENT];
```
Zylann, line 302: `real_t` could be `double`, and this is way too much for a tangent. I think it should be made `float` explicitely.

meshers/blocky/voxel.cpp
```diff
@@ -295,13 +342,20 @@ static void bake_mesh_geometry(Voxel &config, Voxel::BakedData &baked_data) {
 		uvs.resize(positions.size());
 	}
 
+	bool tangents_empty = false;
+	if(tangents.size() == 0) {
+		tangents = PoolVector<real_t>();
```
Zylann, line 347: this assignment does nothing

meshers/blocky/voxel.cpp
```diff
@@ -316,6 +370,23 @@ static void bake_mesh_geometry(Voxel &config, Voxel::BakedData &baked_data) {
 			tri_positions[1] = positions_read[indices_read[i + 1]];
 			tri_positions[2] = positions_read[indices_read[i + 2]];
 
+			float tangent[4];
+
+			if(tangents_empty){
+				//If tangents are empty then we calculate them
```
Zylann, line 376: I think tangents should be optional. Users should be allowed to provide a mesh without tangents if they don't need fancy textures. If they are missing and wanted, it's should be Godot importer/DCC exporter job to fix it.

Ideally, the presence of tangents should be determined per surface, if the material uses normalmaps. Or it would be a boolean on the library, but I'm not sure about the design of this (it's late here).

meshers/cubes/voxel_mesher_cubes.cpp
```diff
@@ -343,6 +343,11 @@ void build_voxel_mesh_as_greedy_cubes(
 					Vector3 n;
 					n[za] = m.side == SIDE_FRONT ? -1 : 1;
 
+					Vector3 t = n.cross(v0);
```
Zylann, line 346: Why add tangents to this mesher? It only makes coloured cubes. It does not even have UVs.

meshers/blocky/voxel.cpp
wrathofrathma: Godot's visual server expects a real array, so I made it one worried I might break it. I'll give this a go.

meshers/blocky/voxel.cpp
wrathofrathma: I agree. The issue is that we'd need some sort of flag to toggle the mesh array in voxel_mesher_blocky.cpp because the visual server checks the packed array to be the exact correct size. Maybe I add a class flag for whether they're present?

meshers/blocky/voxel.cpp
wrathofrathma: I've changed it to that. I dumped the normalized tangents and stored one for each face in the cube_tables.

meshers/blocky/voxel.cpp
wrathofrathma: So I started to implement optional tangents based on whether they're included in the original mesh and if not, just passing an empty array. A problem came to mind that if I do it in a naive manner like this, then if we have a chunk sent to the blocky_mesher that includes a mesh with the tangents, and one without, the visual server will throw a fit because the count doesn't line up. 

I'll think on it some more, but I've made the rest of the requested changes.

meshers/cubes/voxel_mesher_cubes.cpp
wrathofrathma: You're not wrong lol. I think this was just the first place I looked and accidentally mentally linked it to the geometry cube. Removed my changes to this mesher.

meshers/blocky/voxel.cpp
wrathofrathma: I think you're right. Tangents typically follow the direction of the texture to avoid ugly edges and I was worried because the UVs are dependent on the atlas size we'd run into issues. But s will always be positive, so the UVs won't ever flip directions. Which means our normalized tangents should never change? 

I switched to a lookup table in the cube tables file.

meshers/blocky/voxel.cpp
Zylann: Yeah that last problem is what made me unsure, either way it has to be known up front, hence the idea of a boolean option

wrathofrathma: Yeah. I see two possibilities that keep the library instance hidden from the voxel.cpp.

- Expose the toggle in the editor under VoxelLibrary
- Expose the toggle in the editor under Voxel

In both cases we create a method in the voxel library for looping through and setting the flag on existing voxels, and add a few lines to the methods that set/add voxels to the library to check or set the flags in the voxels being added. 

Design-wise in the editor I'm leaning towards under VoxelLibrary because the change is a global change and including it under Voxel doesn't reflect that. We'd just add an option like "Bake Tangents" or something under Voxel Count. 

Do you have any other ideas/what would you prefer?

meshers/blocky/voxel.cpp
```diff
@@ -94,6 +94,10 @@ void Voxel::set_color(Color color) {
 	_color = color;
 }
 
+void Voxel::set_bake_tangents(bool bt) {
+	_bake_tangents = bt;
```
Zylann, line 98: I think the flag doesnt have to exist on `Voxel`, it may be a property of `VoxelLibrary` only. Then it can get passed to `Voxel::bake` functions as an argument.

meshers/blocky/voxel.cpp
```diff
@@ -485,6 +491,8 @@ void Voxel::_bind_methods() {
 
 	ClassDB::bind_method(D_METHOD("is_empty"), &Voxel::is_empty);
 
+	ClassDB::bind_method(D_METHOD("set_bake_tangents", "bake_tangents"), &Voxel::set_bake_tangents);
```
Zylann, line 494: Doesnt need to be here either

meshers/blocky/voxel_library.h
```diff
@@ -86,6 +89,7 @@ class VoxelLibrary : public Resource {
 	std::vector<Ref<Voxel> > _voxel_types;
 	int _atlas_size = 16;
 	bool _needs_baking = true;
+	bool _bake_tangents = false;
```
Zylann, line 92: We can make it default to true as it's more an optimization feature, so it won't confuse users who don't know why shading gets wrong with normal maps

meshers/blocky/voxel.cpp
wrathofrathma: Switched to doing it that way and being on by default.

meshers/blocky/voxel.cpp
```diff
@@ -230,12 +230,20 @@ static void bake_cube_geometry(Voxel &config, Voxel::BakedData &baked_data, int
 		for (unsigned int i = 0; i < 4; ++i) {
 			uvs[i] = (config.get_cube_tile(side) + uv[i]) * s;
 		}
+
+		if(bake_tangents){
+			std::vector<float> &tangents = baked_data.model.side_tangents[side];
+			for (unsigned int i = 0; i < 12; ++i) {
```
Zylann, line 236: Where does the 12 come from?
This is going to add 12 sets of tangents per side, is that wanted?

wrathofrathma: It's been awhile since I looked at this but I'm pretty sure this was intentional. The visual server(visual_server.cpp ~line 500) expects 4x the number of vertices for tangents. The vertex array is received as an array of uint8_t instead of Vector3 so we have to consider the flattened size. 

For each side made up of 4 vertices, there will be 12 floats. So I added one tangent per each float(since the tangent is 4 floats). 

I don't remember if there was any deeper reason based on research. It might honestly be wrong.

Zylann: I thought there should be 4 floats per vertex (3 for tangent, 1 for bitangent?). Which means it would add up to 16 floats per side.
However this `for` is nested. 12 * 4 means 48 floats, and this is already within a third `for` iterating over each side, si it doesn't sound right. I wonder what `i` or `12` even is, maybe it should just not be there?

`VisualServer` deals with `uint8_t` at some point but that's only for compression or data conversion before it sends it to OpenGL, I don't think it's relevant here.

wrathofrathma: Sorry for resurrecting this. Been busy.

I did some testing and I think you're right. When I change it to 4 it works fine. I'll patch it and commit in a bit.

cube_tables.cpp
```diff
@@ -199,4 +199,12 @@ const Vector3i g_ordered_moore_area_3d[MOORE_AREA_3D_COUNT] = {
 	Vector3i(1, 1, 1)
 };
 
+const float g_side_tangents[SIDE_COUNT][4] = {
+	{0.0f, 0.0f, -1.0f, -1.0f},
+	{0.0f, 0.0f, 1.0f, -1.0f},
+	{1.0f, 0.0f, 0.0f, -1.0f},
+	{-1.0f, 0.0f, 0.0f, -1.0f},
+	{1.0f, 0.0f, 0.0f, -1.0f},
```
Zylann, line 207: This array has only 5 elements, but cubes have 6 sides. How curious this compiled for you Oo

wrathofrathma: Uhhhh. Lmao you're not wrong. Apparently I'm writing tangents to 4 of the sides? So it only having 5 isn't breaking it. It's very curious this makes it through the visual server without dying.

I'll fix it and the loop. Starting to feel like I'm losing brain cells with my recent programming.

Zylann: No I think your code is correct, it is really adding to the 6 faces because the outer loop iterates from 0 to `SIDE_COUNT`. The next loop runs 4 times for the 4 vertices, and the innermost loop runs 4 times because each tangent has 4 elements.

Zylann:
I made a few fixes to your PR: https://github.com/Zylann/godot_voxel/commit/1a8cc4380f497709c3e66ac532d81c6ca5819a8c
Notably, tangents were facing the wrong direction for the OpenGL axis convention.

Now it is working:
image

Thanks!

PR #235, opened 2021-03-05
Fix dead loop in VoxelBuffer

great90:
Fix dead loop in 'VoxelBuffer::for_each_voxel_metadata_in_area' and 'VoxelBuffer::clear_voxel_metadata_in_area'


Zylann:
Thanks!

PR #238, opened 2021-03-07
Fix tautological-compare error

stengun:
Comparing the bitwise 'and' result and a number inside the if in this way results always false unless the bitwise 'and' result is 1.

If the intended use of the variable `tests` is as bitflags, then it's wrong to compare the different results to the same integer number since the bitwise & operation can return at most 0 or the bitmask. This triggered a compiler error when compiled with -werror=yes (and curiously this was not triggered in the CI despite the compile command including that flag)


Zylann:
Well spotted! Maybe this didnt get raised because the code isn't actually used anywhere. It's just a manual test that I keep around when I do some research on noises.

PR #240, opened 2021-03-09
bug fix in voxel_lod_terrain.cpp

### LINKED ISSUE #237: Wrong bounds in neighbor checks
Just found a mistake here:
https://github.com/Zylann/godot_voxel/blob/aa0eb670b62987ac77034d6429c348a22a5e74d7/terrain/voxel_lod_terrain.cpp#L615-L617
It should be X, Y and Z, not X all the time.
Same here https://github.com/Zylann/godot_voxel/blob/aa0eb670b62987ac77034d6429c348a22a5e74d7/terrain/voxel_lod_terrain.cpp#L649-L651

Is a problem for non-infinite terrains having non-cubic bounds.

Made this ticket as a reminder to fix this, I found it while working on an experimental branch.

Anshul7sp1:
Attempts to fix #237
Fixes a bug in calculation of bounds for neighbor check.


Zylann:
Thanks!

PR #243, opened 2021-03-12
Update file_locker.h

lhinuz:
Missing file pointer convert.


Zylann:
Thanks! Didn't notice it, as the code using it ended up single-threaded anyways, but it can still cover a few edge cases

PR #248, opened 2021-04-01
Add missing bindings for constants

stengun:
Without those bindings, exported api.json generates the class without the enum despite having a method with a return value of that type. This fixes compilation with generated headers in godot-cpp


Zylann:
Thanks!

PR #249, opened 2021-04-04
Fix instance generator using wrong height for filtering

jacobcoughenour:
The instance generator was using the y position relative to the block when checking if the instance is within the min_height and max_height range. I fixed it by having use the y position relative to the terrain for filtering instead.


Zylann:
Thanks!

PR #251, opened 2021-04-07
Added Mono builds in a new Github Actions workflow

### LINKED ISSUE #244: Provide Mono builds
It would be incredibly useful if the project offered Mono builds as well as standard builds for it's pre-compiled binaries.

I have tried building my own Godot but I can't get the Mono build to build properly locally (msbuild failure).

jacobcoughenour:
This resolves #244 

I added Mono builds for Linux and Windows as a third github actions workflow. 
This works by generating the Mono glue in the first step and then sharing that glue with the Linux and 
Windows steps. This dependency is why I split it off into its own workflow because the alternative would generate the glue twice.

### DIFF
--- a/.github/workflows/mono.yml
+++ b/.github/workflows/mono.yml
@@ -1,4 +1,4 @@
-name: Mono Builds
+name: 🐒 Mono Builds
 on:
   push:
     branches: [ master, github_actions ]
--- a/.github/workflows/mono.yml
+++ b/.github/workflows/mono.yml
@@ -1,0 +1,257 @@
+name: Mono Builds
+on:
+  push:
+    branches: [ master, github_actions ]
+  pull_request:
+    branches: [ master ]
+
+# Global Cache Settings
+env:
+  GODOT_BASE_BRANCH: 3.x
+  SCONS_CACHE_LIMIT: 4096
+
+jobs:
+
+  mono-glue:
+    runs-on: "ubuntu-20.04"
+    name: Generate Mono Glue
+
+    steps:
+      # Clone Godot
+      - uses: actions/checkout@v2
+        with:
+          repository: godotengine/godot
+          ref: 3.x
+
+      # Clone our module under the correct directory
+      - uses: actions/checkout@v2
+        with:
+          path: modules/voxel
+
+      # Azure repositories are not reliable, we need to prevent azure giving us packages.
+      - name: Make apt sources.list use the default Ubuntu repositories
+        run: |
+          sudo rm -f /etc/apt/sources.list.d/*
+          sudo cp -f misc/ci/sources.list /etc/apt/sources.list
+          sudo apt-get update
+
+      # Install all packages (except scons)
+      - name: Configure dependencies
+        run: |
+          sudo apt-get install build-essential pkg-config libx11-dev libxcursor-dev xvfb \
+            libxinerama-dev libgl1-mesa-dev libglu-dev libasound2-dev libpulse-dev libudev-dev libxi-dev libxrandr-dev yasm
+
+      # Upload cache on completion and check it out now
+      - name: Load .scons_cache directory
+        id: mono-glue-cache
+        uses: actions/cache@v2
+        with:
+          path: ${{github.workspace}}/.scons_cache/
+          key: ${{github.job}}-${{env.GODOT_BASE_BRANCH}}-${{github.ref}}-${{github.sha}}
+          restore-keys: |
+            ${{github.job}}-${{env.GODOT_BASE_BRANCH}}-${{github.ref}}-${{github.sha}}
+            ${{github.job}}-${{env.GODOT_BASE_BRANCH}}-${{github.ref}}
+            ${{github.job}}-${{env.GODOT_BASE_BRANCH}}
+
+      # Use python 3.x release (works cross platform; best to keep self contained in it's own step)
+      - name: Set up Python 3.x
+        uses: actions/setup-python@v2
+        with:
+          # Semantic version range syntax or exact version of a Python version
+          python-version: '3.x'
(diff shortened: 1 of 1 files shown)


Zylann:
Is generating the glue twice that big of a problem though? I was more thinking of appending Mono builds to existing scripts since it can re-use the classic builds made earlier to generate the glue (Here we now have 6 builds to make instead of 4)

jacobcoughenour:
So this PR would add 3 more builds of Godot to the CI for a total of 6. Generating the glue twice would result in 7 total builds of Godot but would not rely on artifacts to pass it around. This is without building the mono templates, so adding those would add another 4 Godot builds (2 for glue and 2 for the mono templates).

I'm going to go ahead and try this out anyways so we can compare the two approaches.

Zylann:
Why would generating the glue twice require more builds? Isn't it possible to generate the glue from the already-made classic godot builds?

jacobcoughenour:
> Edit: I checked and indeed it's not possible to use a classic build... kind of sucks.

For future reference:

To build the glue, a temporary Godot has to be built with `module_mono_enabled=yes mono_glue=no` and then it's used generate the glue like this: `<godot_binary> --generate-mono-glue modules/mono/glue`
More info in the docs:
https://docs.godotengine.org/en/stable/development/compiling/compiling_with_mono.html#generate-the-glue

> However I'm told the extra build is supposed to be fast since it can re-used the previous compilation cache

In my [tests on my fork ](https://github.com/jacobcoughenour/godot_voxel/actions/runs/724491402), the glue step takes around 3.5 minutes with a previous compilation cache. So yes, it's not that big of a time difference.

Zylann:
It might be better to re-use the glue then I guess

Zylann:
However I noticed the built Mono version executable is 439 Mb on Windows and 185 Mb on Linux, that doesnt sound right

Zylann:
Still doesnt look right, 146 Mb is way too high for an editor build. I thought `debug_symbols` was also not required, and yet even with it set to `no` the result is too big... usually Mono builds of the editor weight around 76 Mb on Windows, with a sibling folder containing GodotSharp and tools of about 80Mb.

jacobcoughenour:
It turns out the Windows artifact included debug symbols which explains it's size. I disabled them and let the action run again.

So with the artifacts built off my latest commit, here are the sizes:
```
godot.x11.opt.tools.64 92.8 MB
godot.x11.opt.tools.64.mono 185 MB
godot.windows.opt.tools.64.exe 66.4 MB
godot.windows.opt.tools.64.mono.exe 146 MB
```
The artifacts with Mono seem to be double the size of the artifacts without Mono. This still sounded off to me, so I checked the sizes of release v3.2.3 from https://godotengine.org/download:
```
Godot_v3.2.3-stable_x11.64.zip 28.7 MB
Godot_v3.2.3-stable_mono_x11.64.zip 74.7 MB
Godot_v3.2.3-stable_win64.zip 27.5 MB
Godot_v3.2.3-stable_mono_win64.zip 57.9 MB
```
These are release builds so they are smaller but they also double when Mono is included.

I also checked the artifacts from a [recent build from Godot's Actions](https://github.com/godotengine/godot/actions/runs/725591898):
```
linux-template-mono 53.2 MB
linux-editor-mono 170 MB
```

So it looks like the sizes we are getting here are normal.

Zylann:
The size of the executable is still suspiciously high tho https://github.com/Zylann/godot_voxel/pull/251#issuecomment-815093428

Zylann:
Ok sorry the Github Actions page was messing with my mind. The link says `godot.windows.opt.tools.64.mono.exe | 146 MB` but when I click on it it does not actually lead to the exe, but instead a `.zip` which contains the exe separately from other files, and here the sizes check out.

However I'm curious why `debug_symbols` had to be explicitely set to `no` for this to happen? They were not needed for non-Mono builds...

jacobcoughenour:
> However I'm curious why `debug_symbols` had to be explicitely set to `no` for this to happen? They were not needed for non-Mono builds...

The non-Mono builds generate the debug symbols and place them in bin/ but only the .exe is uploaded as the artifact. In the Mono builds, I have it set to upload everything in the bin/ so that it includes the GodotSharp folder but this had the side-effect of also including the debug symbols.

> It should be good for now I think. maybe without the `.exe` at the end of the artifact name
Yeah, I agree.  I removed the .exe now.

Also, the new action could really use an emoji to match the others but I'm not sure which emoji to use.
Any ideas?

Zylann:
Could go for 🪒 or 🧐. There doesn't seem to be an emoji for monocycles

jacobcoughenour:
My multi-lingual friend just informed me that Mono is monkey in Spanish.
So we have 🐒or 🐵 to choose from.  I say we use 🐒 to match the look of 🐧.

Zylann:
Thanks!

Zylann:
The Linux Mono build failed recently and I have no idea what that is https://github.com/Zylann/godot_voxel/runs/2372102603?check_suite_focus=true

jacobcoughenour:
I think something is being missed in the process of uploading and downloading the glue as an artifact. I'll try what we talked about at the beginning, where we have no glue artifact but have to build godot more times, and see if that works.

Zylann:
Turns out the problem was in generating the glue. The job keeps succeeding, even if something fails in generating the bindings :O It appears to be caused by the fact Godot is not exiting with `EXIT_FAILURE` so the job cannot tell if any errors happened.

Zylann:
Hmm, looks like the Mono build workflow just stopped working:
https://github.com/Zylann/godot_voxel/actions/runs/13530677354/job/37811886687
Not sure what is the best fix for this? It says it can't find a `libsndfile` package anymore, but I don't see `libsndfile` anywhere in the workflow file. So perhaps some dependency of something else?

PR #263, opened 2021-05-08
Trying to fix Github build pipeline

FlyveHest:
This seems to be a problem with the Windows images being used to build.

More information in this thread: https://github.com/actions/setup-dotnet/issues/155#issuecomment-761195782


Zylann:
Thanks!

PR #279, opened 2021-06-13
Fix broken documentation links

greenlion:


Zylann:
Seems like the issue comes from a change in MkDocs, so the fix will probably have to be done on the mkdocs config https://github.com/readthedocs/readthedocs.org/issues/8256

Zylann:
Links seem to be back to normal with a22a4d5d3f9ba6ff6c00d85550dae20c915e3126

PR #309, opened 2021-09-14
Added an option in VoxelInstanceGenerator for random rotation

lonevox:
https://user-images.githubusercontent.com/38600896/133188247-0db52f42-ab54-4609-b958-36b47e0890f3.mp4

This option only has a noticeable effect on non-flat terrains when vertical_alignment is 1 (as seen above), because the angle of the terrain determines rotation. This is documented in the code.

### DIFF
--- a/terrain/instancing/voxel_instance_generator.cpp
+++ b/terrain/instancing/voxel_instance_generator.cpp
@@ -286,6 +286,7 @@ void VoxelInstanceGenerator::generate_transforms(
 	const float scale_min = _min_scale;
 	const float scale_range = _max_scale - _min_scale;
 	const bool random_vertical_flip = _random_vertical_flip;
+	const bool random_rotation = _random_rotation;
 	const float offset_along_normal = _offset_along_normal;
 	const float normal_min_y = _min_surface_normal_y;
 	const float normal_max_y = _max_surface_normal_y;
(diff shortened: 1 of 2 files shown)


Zylann:
I'm still curious of what the applications are. What game feature needs this?

lonevox:
I needed it for the visual effect. I have a cube voxel world and wanted pebble details on the ground, but they didn't fit in the art style with random rotation.

Zylann:
Interesting, the terrain types VoxelInstancer supports don't support cubic stuff...

But I did a few experiments with an option to turn off randomization in a more general way, even if the surface is not flat.
The idea is when the option is on, a fixed vector would be used instead of the random one. This essentially makes the instances "look" at the same axis. It mostly works, except when the surface is facing that axis, but not much can be done about that.
If the `up_mode` is `POSITIVE_Y`, the fixed vector would be (1,0,0).
If the mode is `SPHERE`, it would be (0,1,0) instead so that the most discrepancy would be concentrated at the poles.
image

lonevox:
Oh, I didn't even realise it doesn't work with cube voxels... That does make this kinda pointless for what I want to use it for then.

That change you've made does look good though. Better than complete randomness in my eyes.

Zylann:
Closed by c18262d3f89ba1518371af41a8f8939f82d1cf0d

PR #322, opened 2021-10-27
[WIP] Godot 4.0 support based off master

tasgon:
Still a work in progress, doesn't compile yet and there are still quite a few things left to update.

I feel like this would be more appropriate as its own branch, but `godot4` seems to be already taken (and is quite a bit old compared to master, so GitHub's automated diffing tool doesn't take it well).


slapin:
It would be nice to support 3.x version though as it is a lot of time to pass until Godot 4 is usable and a lot until it will pick up supported platforms back (i.e. GLES2) which will make Godot 3.x relevant for quite a while and having supported voxel module would be relevant too.

Zylann:
https://github.com/Zylann/godot_voxel/tree/godot4 is up now. Compiles but runtime bugs may need to be fixed

tasgon:
How did you deal with the meshoptimizer conflicts?

Zylann:
Namespaced it

Zylann:
I think I'll close this PR now because the port was done a while ago. Can still be accessed through the Closed PRs.

PR #326, opened 2021-11-06
Missing includes

alexzk1:
Could not compile as 3 times include <memory> was missing in 3 files for shared_ptr and unique_ptr


Zylann:
Weird. Not sure why it worked locally and on the CI. But when looking at headers without peeking further than STL and Godot headers, I think I can see it missing indeed.

Issue #330, opened 2021-11-11
`debug_print_sdf_top_down` increase memory usage each time when it is called


qarmin:
Fixing this, probably will fix CI fuzzer crashes

Code to reproduce
```
extends Node
# Called when the node enters the scene tree for the first time.
func _process(delta):
	for i in get_children():
		i.queue_free()

######################################## VoxelLodTerrain ########################################
	var temp_variable1 = VoxelLodTerrain.new()
	add_child(temp_variable1)
	temp_variable1.debug_print_sdf_top_down(Vector3(randi()%100,randi()%100,randi()%100),Vector3(randi()%100,randi()%100,randi()%100))
	temp_variable1.debug_print_sdf_top_down(Vector3(randi()%100,randi()%100,randi()%100),Vector3(randi()%100,randi()%100,randi()%100))
	temp_variable1.debug_print_sdf_top_down(Vector3(randi()%100,randi()%100,randi()%100),Vector3(randi()%100,randi()%100,randi()%100))
	
	print("MEM_DYNAMIC: " + str(Performance.get_monitor(Performance.MEMORY_DYNAMIC)/(1024*1024)) + " MB")
	print("MEM_STATIC: " + str(Performance.get_monitor(Performance.MEMORY_STATIC)/(1024*1024)) + " MB")
	print("MEM_DYNAMIC_MAX: " + str(Performance.get_monitor(Performance.MEMORY_DYNAMIC_MAX)/(1024*1024)) + " MB")
	print("MEM_STATIC_MAX: " + str(Performance.get_monitor(Performance.MEMORY_STATIC_MAX)/(1024*1024)) + " MB")
```
Meory will slowly increase from 300 MB to even 2,5GB after few minutes

Zylann:
The reason it happens is because voxel data uses a memory pool. This pool is quite naive at the moment: it assumes most of the time, VoxelBuffer data will have the same sizes. That tends to be true because all chunks and temporary buffers have common sizes, and the rare times a VoxelBuffer is needed with arbitrary size are not often.

When a VoxelBuffer is given a size that was never used before, a new pool is created for blocks of that size. But if you keep using different sizes everytime, it will behave pretty much like a memory leak because none of these allocations will really get re-used...

qarmin:
Looks that `debug_print_sdf_top_down` is the only function that increase memory in this way, so in https://github.com/Zylann/godot_voxel/pull/331 I just disabled it.

[...]

PR #331, opened 2021-11-11
Fix fuzzer freeze and crashes

qarmin:


misc/fuzzer_config.txt
```diff
@@ -0,0 +1,55 @@
+use_parent_methods: true
+add_to_tree: true
+delay_removing_added_nodes_to_next_frame: true
+test_one_class_multiple_times: true
+
+custom_classes:
+FastNoiseLite
```
Zylann, line 7: Note, we could use `config.py` which has a function returning the list of all classes available on the documentation

misc/fuzzer_config.txt
```diff
@@ -0,0 +1,55 @@
...
+VoxelStreamBlockFiles
+VoxelStreamRegionFiles
+VoxelStreamSQLite
+VoxelStreamScript
+VoxelTerrain
+VoxelTool
+VoxelToolBuffer
+VoxelToolLodTerrain
+VoxelToolTerrain
+#VoxelViewer # Cause freeze when using with VoxelTerrain, due iterating each cell
```
Zylann, line 50: What do you mean "iterating each cell"?
Overall, there are some setups that a CI cannot possibly run with a fuzzer. One example is when a fuzzer happens to set LOD count = 1 and a view distance of 1000. This is legit assuming the computer is beefy enough (likely not), but it can bring the machine to its knees.

qarmin: Like I wrote in  #330, running `VoxelTerrain` and `VoxelViewer` freeze when both running(I couldn't find steps to reproduce) due calculating somethings in `process_viewers` function.
Setting `test_one_class_multiple_times` probably should fix this and `VoxelViewer` probably could be uncommented, but I'm not sure in 100%.

Zylann:
I just pushed a rewrite of the memory pool, hopefully allocating buffers of random size repeatedly should no longer increase memory usage over time.

qarmin:
After 10 minutes, memory increase is only ~1/2 MB which is almost identical as in normal Godot functions, so I think that this is fixed already.
Also I see that now with `test_one_class_multiple_times` there is no freeze since `VoxelViewer` and `VoxelTerrain` runs in different frames.

PR #341, opened 2021-12-14
Add missing includes to 3.4 branch

dodopod:
godot3.4 branch won't build without 633e9ab and probably a few other changes

### DIFF
--- a/config.py
+++ b/config.py
@@ -19,6 +19,8 @@ def get_doc_classes():
     "VoxelColorPalette",
     "VoxelInstanceLibrary",
     "VoxelInstanceLibraryItem",
+    "VoxelInstanceLibraryItemBase",
+    "VoxelInstanceLibrarySceneItem",
     "VoxelInstanceGenerator",
 
     "VoxelBuffer",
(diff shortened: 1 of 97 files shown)


Zylann:
It seems your PR has not been rebased, it carries a lot more commits than it should

dodopod:
Oh, I merged instead of rebasing. Should i open another PR?

Zylann:
You could rebase and then force-push on your branch, or make a new one

Zylann:
Actually, I'm not sure if the 3.4 branch should get the new things present in master. That branch is a feature freeze, it should not get the stuff in master, which may have bugs also. Only bug fixes should be ported there.

dodopod:
All right, I'll change it to only include the commits I need to avoid a compile error.

dodopod:
I've rebased my fork with only the commits I need to build Godot with this module, and changed the title accordingly.

Zylann:
Your commits still seem to contain the addition of `_b_set_node_name`

dodopod:
Is this better?

Zylann:
Yup, thanks!
Note: this will need to be ported to the godot4 banch if you need it as well

Issue #340, opened 2021-12-12
Debug editor support broken with C#/Mono & Mono Builds not working


Gamerfiend:
As referenced in #332, mono builds for the module are currently broken. After many attempts I have been unable to find a solution to build them successfully in either github-actions or locally with the module.

If you go back far enough, you can find builds that are working; Specifically the last known working commit with Mono support is this commit https://github.com/Zylann/godot_voxel/tree/454d8b2c7f4a9d1dd26eefb3f06185e0ad39879e.

Even still, when the builds were working, the mono support itself was lacking when creating a custom script for voxel generation using C#. When running a project in editor, with or without debugging, you will run into an error complaining about `!tls`. Official Godot mono builds get around this error by applying a patch, and building mono. More about that can be seen here, https://github.com/godotengine/godot-mono-builds; with the specific patch being here https://github.com/godotengine/godot-mono-builds/blob/master/files/patches/mono-dbg-agent-clear-tls-instead-of-abort.diff.

However the godot-mono-builds repo doesn't successfully output `.dll` files to be used in windows builds, an issue for that is open here. https://github.com/godotengine/godot-mono-builds/issues/54.

Zylann:
Related https://github.com/Zylann/godot_voxel/issues/332

Gamerfiend:
Just a little update.. I’ve now been trying to build Godot using mingw on windows, and the `godot-mono-builds`, which does allow the `.a` files to be used instead of `.dll` however I have been unsuccessful in my attempts to build it.

Gamerfiend:
Fix for this in https://github.com/Zylann/godot_voxel/pull/344

Gamerfiend:
This issue has been resolved, closing this issue.

PR #344, opened 2022-01-07
Fix for Mono builds

Gamerfiend:
Fixes: https://github.com/Zylann/godot_voxel/issues/340


Gamerfiend:
I suggest squash merging…

PR #345, opened 2022-01-08
Update mono builds to be working, for godot 3.x or 3.4

### LINKED ISSUE #340: Debug editor support broken with C#/Mono & Mono Builds not working

Gamerfiend:
Fixes #340

### DIFF
--- a/.github/workflows/mono.yml
+++ b/.github/workflows/mono.yml
@@ -1,7 +1,7 @@
 name: 🐒 Mono Builds
 on:
   push:
-    branches: [ godot3.x, godot3.4, github_actions ]
+    branches: [ master, godot3.x, godot3.4, github_actions ]
   pull_request:
     branches: [ godot3.x, godot3.4 ]
 
--- a/.github/workflows/mono.yml
+++ b/.github/workflows/mono.yml
@@ -1,46 +1,176 @@
 name: 🐒 Mono Builds
 on:
   push:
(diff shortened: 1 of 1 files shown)


Gamerfiend:
Fixes #340

Zylann:
Why is this PR said to make Godot 3.4 and 3.x work, when you made it target `master`? `master` is now Godot4

PR #346, opened 2022-01-08
Mono CICD for Godot 3.X

Gamerfiend:


Zylann:
If this job is for `godot3.x`, why does it runs on `master`, `godot3.4` and `godot3.x`?

Zylann:
Thanks!

PR #347, opened 2022-01-08
Mono CICD for Godot 3.4

Gamerfiend:


Zylann:
If this job is for 3.4, why does it runs on `master`, `godot3.4` and `godot3.x`?

Zylann:
Thanks!

PR #348, opened 2022-01-16
Misleading documentation

harrisonvanderbyl:
I have no idea how this is supposed to look, but perhaps something like this


Zylann:
Unfortunately this is not something you are supposed to write yourself, this part of the documentation is generated by Godot's documentation tool. Either it needs to be re-run, or Godot has a bug that made it output the wrong return value (which means next time the doc is updated your fix will be overwritten)

PR #349, opened 2022-01-16
This seems broken

harrisonvanderbyl:
I may not know much about the code base, but I'm pretty sure that's a premature exit right there


Zylann:
I made a fix myself https://github.com/Zylann/godot_voxel/commit/8031800785b0868e069fc19a9ec75d56719cf387

PR #375, opened 2022-04-28
Docs: Flip normals

jomiq:
Small doc fix, the dx and dy order was reversed.


Zylann:
Are you sure?
image

jomiq:
Ok. That’s really weird. Let’s keep an eye on it and see what’s really going on. Close this  if you like.

jomiq:
That’s probably it then. ([From the horse’s mouth](https://www.khronos.org/news/permalink/handling-differences-between-vulkan-and-opengl-coordinate-system))

I guess this should be stable in the shader language first, closing PR

Zylann:
@poikilos I'm not sure winding order matters here, it's about a fragment shader, not normals from mesh

PR #380, opened 2022-05-09
Missing imports

kubecz3k:
On my machine (ubuntu 22.04) I was unable to compile the project.

First I was thinking I have some problem with matching proper godot engine version with proper voxel commit (especially that I had no problem with building 3.x version), but then I realized automatic builds are working here so I took the exact commits from two repos and I still had build errors, that made me look more closely at errors and I realized that most probably some of my system libraries were build in a bit different way and do not include some of the imports. 

After adding those everything builds nicely on current masters (godot 943b50995292d98d9bc2e45ff04eaf0a716cea28) (voxel module 668fedfa528f9971577f7bf19041f948db512d3d)

### DIFF
--- a/util/span.h
+++ b/util/span.h
@@ -3,6 +3,7 @@
 
 #include "fixed_array.h"
 #include <vector>
+#include <cstddef>
 
 namespace zylann {
 


meshers/transvoxel/transvoxel.h
```diff
@@ -9,6 +9,7 @@
 
 #include <core/math/color.h>
 #include <vector>
+#include <stddef.h>
```
Zylann, line 12: Was it required here? I think `voxel_buffer_internal.h` already includes `Span`

kubecz3k: Looks you are right, it still compiles without that import :)
Updated the pr

Zylann:
Thanks!

PR #382, opened 2022-05-13
Fixed issues caused by changes to HashMap in godot

Trey2k:
Just a quick fix for changes to hashmaps in the master godot branch https://github.com/godotengine/godot/commit/8b7c7f5a753b43cec10f72b274bb1d70c253652b

### DIFF
--- a/terrain/fixed_lod/voxel_terrain.cpp
+++ b/terrain/fixed_lod/voxel_terrain.cpp
@@ -377,9 +377,10 @@ void VoxelTerrain::set_block_enter_notification_enabled(bool enable) {
 
 	if (enable == false) {
 		const Vector3i *key = nullptr;
-		while ((key = _loading_blocks.next(key))) {
-			LoadingBlock *lb = _loading_blocks.getptr(*key);
-			CRASH_COND(lb == nullptr);
+		
+		for (auto elm = _loading_blocks.begin(); elm < _loading_blocks.end(); ++elm) {
+			LoadingBlock *lb = _loading_blocks.getptr(elm->key);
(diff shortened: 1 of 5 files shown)


terrain/fixed_lod/voxel_terrain.cpp
```diff
@@ -377,8 +377,9 @@ void VoxelTerrain::set_block_enter_notification_enabled(bool enable) {
 
 	if (enable == false) {
 		const Vector3i *key = nullptr;
```
Zylann, line 379: Key is now unused

terrain/fixed_lod/voxel_terrain.cpp
```diff
@@ -377,8 +377,9 @@ void VoxelTerrain::set_block_enter_notification_enabled(bool enable) {
...
-		while ((key = _loading_blocks.next(key))) {
-			LoadingBlock *lb = _loading_blocks.getptr(*key);
+		
+		for (auto elm = _loading_blocks.begin(); elm < _loading_blocks.end(); ++elm) {
+			LoadingBlock *lb = _loading_blocks.getptr(elm->key);
```
Zylann, line 382: Can you use the iterator instead of doing a second hashmap lookup?

Zylann (changes requested):

terrain/fixed_lod/voxel_terrain.cpp
Trey2k: Just pushed changes. Sorry did even noticed I did a redundant lookup.

Zylann:
Thanks!

PR #384, opened 2022-05-13
DOC: Make it clearer where to look on building mismatch

duchainer:
When the latest godot commit and the latest godot_voxel commit don't match, look at the latest successful Github action.

Also the godot_voxel/master is targeting the tip of the godot/master, so with breaking changes, a recent voxel/master commit won't work with all recent godot/master commits.

Still a first draft of what could be said, I'll try sleeping on it to come up with something clearer or at a better place in the doc


Zylann:
In fact it is even true without Godot being in alpha. The master branch will always be expected to build at some master version of Godot, but might not always work with it (which is especially true with alphas).

When it comes to version branches, these are known to work with the latest stable version of Godot known at the time.

Issue #387, opened 2022-05-18
Linking error with Manjaro build of master branch


kilon:
Seems the linker is not happy with building the tool on Manjaro with latest commit 008c38bb6be63858b484101a3b22f1d6b840599b
I tried also a previous commit that reports no build errors for CI 602ae1ea955a9165d728075c1f73944f9e5bac5f but i get the same error. Anyone else experience this ? 

I dont experience any compilers errors, I building with the latest commit of Godot 4 master branch

I paste the error bellow 

```
/usr/bin/ld: modules/libmodule_voxel.linuxbsd.opt.tools.64.a(voxel_blocky_model.linuxbsd.opt.tools.64.o): warning: relocation against `_ZN6zylann5voxel16VoxelBlockyModel9BakedData5Model12MAX_SURFACESE' in read-only section `.text'
/usr/bin/ld: modules/libmodule_voxel.linuxbsd.opt.tools.64.a(voxel_blocky_model.linuxbsd.opt.tools.64.o): in function `zylann::voxel::bake_mesh_geometry(zylann::voxel::VoxelBlockyModel&, zylann::voxel::VoxelBlockyModel::BakedData&, bool, zylann::voxel::VoxelBlockyModel::MaterialIndexer&)':
/home/kilon/coding/kyklops/modules/voxel/meshers/blocky/voxel_blocky_model.cpp:355: undefined reference to `zylann::voxel::VoxelBlockyModel::BakedData::Model::MAX_SURFACES'
/usr/bin/ld: warning: creating DT_TEXTREL in a PIE
collect2: error: ld returned 1 exit status
scons: *** [bin/godot.linuxbsd.opt.tools.64] Error 1
scons: building terminated because of errors.
[Time elapsed: 00:00:41.082]
mv: cannot stat './bin/godot.linuxbsd.opt.tools.64': No such file or directory
```

MGilleronFJ:
First time I ever saw this in my life... no idea what the fix is supposed to be. `MAX_SURFACES` is a constant I defined the same way as many other constants in this module.

duchainer:
Did the fix with @CarlosRios, I'll be opening a PR in a few.
Thanks a lot @MGilleronFJ for the suggestion, this fixed it on our Mac Builds :smile:

Zylann:
Fix merged in f9690542b2defc6a8c60e3456d47609508b72a37

PR #389, opened 2022-05-18
Cross-compiler compatibility: const -> constexpr

duchainer:
As suggested in https://github.com/Zylann/godot_voxel/issues/387
This allowed 2 full Mac builds using these commands:
 ```
scons platform=osx arch=x86_64 --jobs=$(sysctl -n hw.logicalcpu) warnings=all tools=yes tests=no target=release_debug production=yes
 ```
```
scons platform=osx arch=arm64 --jobs=$(sysctl -n hw.logicalcpu) warnings=all tools=yes tests=no target=release_debug production=yes voxel_fast_noise_2=false 
```
 -

( Also after having the VulkanSDK like mentioned here[1] and after deleting the `denoise` module which doesn't like ARM64)

[1]: https://github.com/godotengine/godot/issues/57622#issuecomment-1120486153

### DIFF
--- a/meshers/blocky/voxel_blocky_model.h
+++ b/meshers/blocky/voxel_blocky_model.h
@@ -63,7 +63,7 @@ public:
 		};
 
 		struct Model {
-			static const uint32_t MAX_SURFACES = 2;
+			static constexpr uint32_t MAX_SURFACES = 2;
 
 			// A model can have up to 2 materials.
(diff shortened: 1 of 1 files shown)


duchainer:
Let me know if I missed anything, I don't have much C++ experience so I'm still at the tinkering phase :sweat_smile:

Zylann:
Thanks!

PR #395, opened 2022-05-26
Avoid unused variable warning.

rcorre:
gcc warns:

```
modules/voxel/streams/region/region_file.cpp: In function 'bool save_header(FileAccess*, uint8_t, const VoxelRegionFormat&, const std::vector<VoxelRegionBlockInfo>&)':
832
modules/voxel/streams/region/region_file.cpp:101:9: error: unused variable 'blocks_begin_offset' [-Werror=unused-variable]
833
  101 |  size_t blocks_begin_offset = f->get_position();

	const size_t blocks_begin_offset = f.get_position();
```

I noticed this while trying to build export templates using github
actions, as the godot actions treat warnings as errors by default:

https://github.com/rcorre/godot/runs/6617207436


Zylann:
Thanks!

Issue #397, opened 2022-06-01
The test project test seems wrong


poikilos:
The Test Project output (See [runs](https://github.com/Zylann/godot_voxel/actions/workflows/test_project.yml) of the test) appears to have a problem related to case sensitivity:
```
  inflating: regression-test-project-3.x/project.godot  
mv: cannot stat 'RegressionTestProject-3.x': No such file or directory
```

The archive is from github.com/garmin but don't see the word "garmin" anywhere in [the .github/test_project](https://github.com/Zylann/godot_voxel/blob/master/.github/workflows/test_project.yml) so I wouldn't know where to start fixing it. Maybe the issue is in something that the test downloads before it downloads the file being inflated in the quote. The script in question expects "RegressionTestProject-3.x" but the folder is actually called "regression-test-project-3.x" so it is a case sensitivity issue.

The following branches don't have the issue:
- godot3.1
- godot3.2.3
- godot3.3

After the following branches are fixed, the issue could could be closed:
- [x] https://github.com/Zylann/godot_voxel/pull/399 (godot3.x)
- [x] https://github.com/Zylann/godot_voxel/pull/398 (master (godot4))
- [x] https://github.com/Zylann/godot_voxel/pull/402

Zylann:
> The archive is from github.com/garmin

It's `qarmin`, with a Q

poikilos:
Thanks, the accepted pull requests seem to cover this. If there are other branches you can reopen it and I can work on them.

PR #398, opened 2022-06-03
Use the new project name and avoid a URL redirect.

poikilos:
Related to issue #397:
- Use the updated URL (I used the URL to which the old URL redirects).
- Use the updated project name (since an invalid source in `mv` command was causing a test to fail on all commits).

This PR only fixes master (I will also look at 3.x branch and add another PR as necessary).

### DIFF
--- a/.github/workflows/test_project.yml
+++ b/.github/workflows/test_project.yml
@@ -80,9 +80,9 @@ jobs:
       # Download and extract zip archive with project, folder is renamed to be able to easy change used project
       - name: Download test project
         run: |
-          wget https://github.com/godotengine/regression-test-project/archive/master.zip
-          unzip master.zip
-          mv "regression-test-project-3.x" "test_project"
+          wget https://github.com/godotengine/regression-test-project/archive/refs/heads/4.0.zip
+          unzip 4.0.zip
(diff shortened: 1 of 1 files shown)


poikilos:
I'm not sure whether you want this on the master branch or you want to cherry pick from godot3.x branch later so I'll add another PR for that branch and you can handle this as you wish.

poikilos:
Ah, good point. Their default branch is 3.x, and using that implicitly isn't desirable. I left it that way since that's how the test was before. I'm not sure what Godot version your master branch is targeting, so I'd have to adjust if you can answer that.

Zylann:
The `master` branch is targetting Godot's `master` branch, which is basically 4.0

Zylann:
I wonder why the test project fails to compile the engine when the linux and windows jobs succeed

poikilos:
The tests have been failing 100% of the time already: https://github.com/Zylann/godot_voxel/actions/workflows/mono.yml
That's why I thought fixing the test would help out, to separate signal from noise so the real errors can be seen.

Zylann:
You linked the mono builds, these don't run the same workflow.
I'll merge anyways because it may help after compilation is fixed (although I'm not sure yet what needs to be fixed)

PR #401, opened 2022-06-13
Add `*.tmp` to the .gitignore

aaronfranke:
When compiling, I got temporary files showing up in my Git client with the extension `.tmp`. This PR ignores them. I also alphabetized the list so it's easy to find things.


MGilleronFJ:
Do you know where they come from?

PR #402, opened 2022-06-13
Account for the renamed test project in the godot3.4 branch (resolves #397).

### LINKED ISSUE #397: The test project test seems wrong

poikilos:
This appears to fix the last branch affected by issue #397 so I made a list there for clarity. I only checked versioned branches and master, but if any other branch is active please comment and I can fix it.

### DIFF
--- a/.github/workflows/test_project.yml
+++ b/.github/workflows/test_project.yml
@@ -80,9 +80,9 @@ jobs:
       # Download and extract zip archive with project, folder is renamed to be able to easy change used project
       - name: Download test project
         run: |
-          wget https://github.com/qarmin/RegressionTestProject/archive/3.x.zip
+          wget https://github.com/godotengine/regression-test-project/archive/3.x.zip
           unzip 3.x.zip
-          mv "RegressionTestProject-3.x" "test_project"
+          mv "regression-test-project-3.x" "test_project"
 
       # Editor is quite complicated piece of software, so it is easy to introduce bug here
       - name: Open and close editor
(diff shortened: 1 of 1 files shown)


Zylann:
There is no archive for Godot 3.4? Note that 3.4 is a feature freeze, if you use 3.x it will test against a more recent Godot version, which might not be compatible.

poikilos:
Good catch, I've switched to the 3.4 regression test.

PR #407, opened 2022-07-04
Optimize SVG images and convert most PNG images to WebP

aaronfranke:
This PR losslessly optimizes PNG and SVG images using `oxipng` and `svgcleaner`, and converts most PNG images to WebP. The images are the same pixels, they just have a smaller file size.

I also added that pesky `.DS_Store` to the `.gitignore` for macOS.


Zylann:
Thanks!

PR #410, opened 2022-07-14
implemented flatten in the voxel_tool

jeronimo-schreyer:


MGilleronFJ:
I might merge it as an initial implementation, but it can be implemented in a much more optimal way, because `_set_voxel` is the slowest way to modify voxels.
Also, I dont think it's worth supporting every "mode" somehow since flattening only makes sense with removing matter.
This implementation also isn't smooth and causes sharp discontinuities in SDF, but I guess it works as a first step.

Zylann:
I added `do_hemisphere` if you want to give it a try

PR #419, opened 2022-08-03
Run doctool to update docs for Godot 4

aaronfranke:
This PR runs doctool to update the docs for Godot 4. I looked through to check that no still-valid descriptions are deleted. There are still many descriptions that get deleted, but only because the classes have changed a lot.


Zylann:
Thanks!

PR #420, opened 2022-08-04
Rename `flush_cache` with connection parameter to `flush_cache_to_connection`

aaronfranke:
Also rename the argument to `p_connection`.


MGilleronFJ:
The name would imply it is the cache of the connection, however there is only one cache, so perhaps it should be called `flush_cache_to_connection`? SQLite supports multiple connections to the same database, which is why there is a pool of them for multithreading, but I might change that in the future because the way SQLite does multithreading isn't good for the use case.

PR #435, opened 2022-08-26
`deg2rad` was renamed in Godot (godotengine/godot#64367)

menip:
I checked the other renames as well, this was the only one.


Zylann:
Thanks!

PR #436, opened 2022-08-27
Run doctool, use "param" instead of "argument"

aaronfranke:
In the XML docs, the "argument" tag was renamed to "param". This PR runs doctool to update this. Running doctool also added and removed a few things from the XML docs. The parameter descriptions in `VoxelBlockyLibrary.xml` was not valid since parameter documentation is currently not supported in Godot's documentation XML.

### DIFF
--- a/doc/classes/VoxelBlockSerializer.xml
+++ b/doc/classes/VoxelBlockSerializer.xml
@@ -27,19 +27,19 @@
 	<methods>
 		<method name="deserialize">
 			<return type="void" />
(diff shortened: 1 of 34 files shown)


Zylann:
Ah, didnt see the `/>`. Can the descriptions be moved to the function description?

Also I realize some functions were bound with missing parameters, but Godot didn't warn at all and happily filled in the parameters in docs as "_unnamed_arg0" or "_unnamed_arg1... I should fix them too

Zylann:
Thanks!

PR #443, opened 2022-09-23
Update get_configuration_warnings to use PackedStringArray

aaronfranke:
This PR allows the Voxel module to compile on the latest Godot master branch.


MGilleronFJ:
This has also been done in the `gdextension` branch. Maybe I should merge that one now.

aaronfranke:
We are still using the module for now, it's likely worth keeping it maintained. Maybe on a separate branch. The maintenance would mostly be to keep it compiling and then new features can go on the GDExtension branch.

MGilleronFJ:
The gdextension branch doesn't remove support for compiling as a module, it still works the same.

PR #447, opened 2022-10-08
Misc fixes to includes and build system

aaronfranke:
* Fixes to includes to make CI builds not compile with errors.
* Fix `find_godot` function in `doc/tools/build.py` and add support for macOS and other architectures (I didn't check if the whole script works but I did check that this specific file name logic and the regex is correct).
* Remove commented-out unit test running code in the workflows (the TODO comment was confusing and the file path is no longer correct, so remove and they can just be added back later if needed).


Zylann:
Thanks!

PR #459, opened 2022-10-31
Re-enable fuzzer

qarmin:
This PR re-enable fuzzer which after change to Godot 4.0 was broken(master branch use Vulkan and without SwiftShader it fails silently)

I changed `util/godot/node.h` because I cannot compile it on my computer.

Now fuzzer contains also reproducer - small program that automatically will try to find the smallest possible amount of code that crashes/leaks/timeouts app.

I don't think that Godot regression project is needed anymore(It was also broken due brotolli output messages that triggered CI failure)

Rest CI jobs I will restore when Fuzzer job will pass tests


Zylann:
Your PR is deleting most of the CI scripts (windows, extension_windows, linux, mono), I dont think that is intented.

What is the `suppressions` file for?

qarmin:
To test this fuzzer CI I only need one workflow file, so I removed other to not spam unrelated jobs until I done.
I amended some commits, so maybe that is why workflows shows different times.
Suppression file is for undefined sanitizer messages that are visible when opening editor/running project with Godot(of course when using undefined sanitizer) like:
```
thirdparty/harfbuzz/src/hb-open-type.hh:624:18: runtime error: index 2 out of bounds for type 'Index [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:624:18: runtime error: index 2 out of bounds for type 'Index [1]'
thirdparty/icu4c/common/rbbi.cpp:904:32: runtime error: index 11 out of bounds for type 'unsigned char [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:624:18: runtime error: index 2 out of bounds for type 'OffsetTo [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:624:18: runtime error: index 2 out of bounds for type 'OffsetTo [1]'
thirdparty/harfbuzz/src/hb-open-type.hh:624:18: runtime error: index 3 out of bounds for type 'Record [1]'
```
which comes usually from thirdparty libraries so are hard to fix or just false positives - https://github.com/harfbuzz/harfbuzz/issues/2953, so with suppression file, output of fuzzer/reproducer should not contains this messages
Screenshot from 2022-10-31 21-17-54

Zylann:
I just merged a branch into master, fixed reported crashes and fixed `queue_free`.

Zylann:
Is there something more pending? I see the scripts are still in deleted state.

qarmin:
Looks that reproducer fails to run, because I messed with some paths, so I need to fix them to be able to restore other CI jobs
```
/usr/bin/xvfb-run: 184: ../bin/godot.linuxbsd.editor.dev.x86_64.san: not found
```
After that CI should probably will fail with https://github.com/Zylann/godot_voxel/issues/461 which will mean that everything works.

MGilleronFJ:
Looks like it failed, but it says:
```
ERROR: Assertion failed in project, check execution log for more info (search for "Assertion failed" in CI log)
```
I don't know where to look actually, because I dont only use Godot error macros, I also use custom ones that work in reverse, and they log "Assertion failed" even in cases that do not cause a crash... however, the ones that crash do log "FATAL" instead. I couldn't find any.

qarmin:
Removed check for assertion, since in this project such message happens very often and is not fatal/critical.

To simplify finding bugs, this is how looks action tab
Screenshot from 2022-11-06 10-04-36

- godot-linux - compiled binary with sanitizers which can be used to locally reproduce bugs
- last-run-function - list of functions executed in last project frame
- project-results - 200 lasts lines from each fuzzer run to see with which messages Godot closes/crashes
- reproducer_* - minimal code which should crash, leaks or timeouts found by reproducer(not always available and sometimes may contains unnecessary lines, but in 95% code is shorter than 10 lines)

also entire last run - output, crash and code to reproduce, can be found inside `Run reproducer` steps inside CI
Screenshot 2022-11-06 at 10-12-32 🐺️ Fuzzer · qarmin_godot_voxel@1571e6b

MGilleronFJ:
I feel like having 2.32 gb being uploaded on every run is kinda wasted because I don't work on Linux, but mainly I wonder if that will trigger Github storage limits? I'm currently using regular builds of the CI to provide executables to users who dont want to compile Godot.

qarmin:
I store 10 GB artifacts for 14 days, and I don't have any problems with limits - https://github.com/qarmin/GodotBuilds/actions/runs/3406401661.
The only way to remove this artifact is to unify editor and fuzzer step, which I separated to easily restart step with fuzzer when needed(without it Godot will need to build again - looks that with address sanitizer this takes 20-30 minutes)

.github/workflows/fuzzer.yml
```diff
@@ -68,62 +54,206 @@ jobs:
...
-      # Use python 3.x release (works cross platform; best to keep self contained in it's own step)
-      - name: Set up Python 3.x
-        uses: actions/setup-python@v2
+      # NO-PIE is for more detailed bactraces with line symbols
+      - name: Compilation
+        env:
+          SCONS_CACHE: ${{github.workspace}}/.scons_cache/
+        run: |
+          # Helpful to test later code without needing to compile entire editor
+          #wget https://downloads.tuxfamily.org/godotengine/4.0/beta3/Godot_v4.0-beta3_linux.x86_64.zip -O a.zip
```
Zylann, line 63: beta4 was released since. This line will have to be updated regularly, not sure if there is an alternative

Zylann:
Thanks!

Zylann:
This fuzzer job found a crash, but there doesn't seem to be artifacts available, is it expected? https://github.com/Zylann/godot_voxel/actions/runs/3465267839/jobs/5788624072

qarmin:
Sometimes reproducer is not able to found crash/leak when happens random due thread races or similar things, but this time properly prepared artifacts, which are available in summary page.

[Nagranie ekranu z 2022-11-15 07-33-39.webm](https://user-images.githubusercontent.com/41945903/201844767-1393b0b6-d367-45ed-8360-68d49953fbc8.webm)

Zylann:
Aaah I see, I keep missing the spot^^"

PR #475, opened 2022-12-20
Fix compile error in GPU storage buffer pool debug print

aaronfranke:
The return type of `pool.buffers.size()` has an ambiguous conversion to Variant, at least on Clang on macOS (since the CI succeeds for Windows and Linux I guess this may be a Clang or Mac specific issue).

This PR explicitly converts them to `int64_t` before using them in the Variant array. I used `int64_t` because that's the type Variant stores internally, so it's the least amount of casting, but the conversion would also work using `uint64_t` or `unsigned int` or any other type Variant is able to be constructed from.


Zylann:
Ah, this one may be fixed with `ZN_SIZE_T_TO_VARIANT` I think

PR #479, opened 2023-01-11
Fix compiling with Godot's MouseButton mask bitfield changes (beta 11)

aaronfranke:
See https://github.com/godotengine/godot/pull/71045 for the engine and https://github.com/godotengine/godot-cpp/blob/master/gdextension/extension_api.json for GDExtension.

Should work in both the engine module and GDExtension but I only tested as an engine module.

### DIFF
--- a/util/godot/core/mouse_button.h
+++ b/util/godot/core/mouse_button.h
@@ -5,18 +5,18 @@
 #include <core/input/input_enums.h>
 
 #define ZN_GODOT_MouseButton_NONE MouseButton::NONE
-#define ZN_GODOT_MouseButton_MASK_MIDDLE MouseButton::MASK_MIDDLE
 #define ZN_GODOT_MouseButton_WHEEL_UP MouseButton::WHEEL_UP
 #define ZN_GODOT_MouseButton_WHEEL_DOWN MouseButton::WHEEL_DOWN
+#define ZN_GODOT_MouseButtonMask_MIDDLE MouseButtonMask::MIDDLE
 
 #elif defined(ZN_GODOT_EXTENSION)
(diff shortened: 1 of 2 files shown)


util/godot/core/mouse_button.h
```diff
@@ -5,7 +5,7 @@
 #include <core/input/input_enums.h>
 
 #define ZN_GODOT_MouseButton_NONE MouseButton::NONE
-#define ZN_GODOT_MouseButton_MASK_MIDDLE MouseButton::MASK_MIDDLE
+#define ZN_GODOT_MouseButton_MASK_MIDDLE MouseButtonMask::MIDDLE
```
MGilleronFJ, line 8: Perhams "MASK" should be removed from the name

util/godot/core/mouse_button.h
```diff
@@ -14,7 +14,7 @@
 using namespace godot;
 
 #define ZN_GODOT_MouseButton_NONE MouseButton::MOUSE_BUTTON_NONE
-#define ZN_GODOT_MouseButton_MASK_MIDDLE MouseButton::MOUSE_BUTTON_MASK_MIDDLE
+#define ZN_GODOT_MouseButton_MASK_MIDDLE MouseButtonMask::MOUSE_BUTTON_MASK_MIDDLE
```
MGilleronFJ, line 17: If GDExtension still requires using masks, the code that checks the mouse middle button would have to change, which further complicates this abstraction...

aaronfranke: BitField seems to be already in GDExtension, it's in the API JSON.

PR #483, opened 2023-02-02
Change the type of the parameter "p_usage" of parse_property() to Bit…

SneaK1ng:
…Field<PropertyUsageFlags>.

see:https://github.com/godotengine/godot/commit/c7f4ca36a4541686d5944f7bf88e1aa7d32cc24f

### DIFF
--- a/.github/workflows/extension_windows.yml
+++ b/.github/workflows/extension_windows.yml
@@ -1,7 +1,7 @@
 name: 🚪 Windows Extension Builds
 on:
   push:
-    branches: [ master ]
+    branches: [ master, github_actions ]
   pull_request:
     branches: [ master ]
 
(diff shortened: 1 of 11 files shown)


Zylann:
I need to check if your PR works when building the module as a GDExtension, but Github doesnt seem to allow retriggering jobs with the latest commits and your PR...

SneaK1ng:
Sorry, I will create a new one.

SneaK1ng:
**in editor_inspector_plugin.hpp line 64:**
`virtual bool _parse_property(Object *object, Variant::Type type, const String &name, PropertyHint hint_type, const String &hint_string, BitField<PropertyUsageFlags> usage_flags, bool wide);`

PR #493, opened 2023-03-18
Fix "Failed to load .NET runtime" in mono CI builds

Treer:
Builds [the managed libraries via build_assemblies.py](https://docs.godotengine.org/en/latest/contributing/development/compiling/compiling_with_dotnet.html#building-the-managed-libraries), which adds the `GodotSharp` directory, which fixes this error when starting the editor:
image

(The official release doesn't need the `GodotSharp` directory & files, but I think that's just because they packaged it into the executable with fancier build actions)

Mono is no longer installed on the `Windows Editor w/ Mono` image since the "windows-latest" base comes with dotnet 6.0, which Godot 4 has switched to.

----

* I've added the same `build_assemblies.py` change to the mono-linux build, it builds successfully but I haven't tested the binaries as I don't have a non-remote linux box.
* On Windows, the Godot 4.1.dev Editor binaries built from this action can compile and run my dotnet 6.0 C#-based project. My project doesn't use godot_voxel yet but I can see that nodes like VoxelLodTerrain are available from the Add Node dialog, so for windows at least I believe these editor binaries are working!
* Export Templates are still not supported (the Mono Builds action didn't build them before and I've not attempted to).

### DIFF
--- a/.github/workflows/mono.yml
+++ b/.github/workflows/mono.yml
@@ -77,7 +77,7 @@ jobs:
 
       - name: Generate Mono Glue
         run: |
-          xvfb-run ./bin/godot.linuxbsd.editor.x86_64.mono --generate-mono-glue modules/mono/glue || true
+          xvfb-run ./bin/godot.linuxbsd.editor.x86_64.mono --headless --generate-mono-glue modules/mono/glue || true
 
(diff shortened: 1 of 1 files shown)


Zylann:
Thanks!

PR #503, opened 2023-04-15
New do_blend_ball() function

Piratux:
Added new `do_blend_ball()` function for VoxelTool that smooths out terrain using box blur.

Showcase of this feature:
https://user-images.githubusercontent.com/58703216/232250883-bc55a7b0-f2dd-44b5-b9d3-d55088fa3f6c.mp4


Zylann:
I implemented a similar algorithm in bf8ce2b186d4b89b59bb0bdb2b53fead64a52451. Implemented it in a more self-contained way without using single-voxel get/sets, which are slower than doing a bulk read/write. There is still room for optimization (and ultimately can use SIMD), but it's quite fast.
The only thing I didn't do is to apply an easing to the SDF, not sure why that would be needed.

Piratux:
The easing was left over from my 2D experiments of blending (which produces better visually looking results).

For 3D this extra processing step could be made linear to match `_sdf_scale` which could look something like this:
```cpp
float tmp_sd = tmp[tmp_loc];

// Function that returns value greater than 0.
// Memory optimisation would occur when function returns value greater than 1 as that's when unwanted precision would be cut off.
// However I can't think of a function that would map relation between _sdf_scale and averaged sdf multiplier to achieve this effect.
tmp_sd *= function_that_depends_on_sdf_scale(_sdf_scale);

const float sd = Math::lerp(src_sd, tmp_sd, factor);
```
This could serve as possible memory usage optimisation (considering values with -1 or 1 in a chunk are stored more efficiently) as this way only voxels that are close to surface would have values in interval (-1; 1).
However, I don't have a strong opinion on whether this should be added.

Piratux:
Since requested feature was added to the module as VoxelTool.smooth_sphere(..) I'll close this.

PR #525, opened 2023-06-07
Update third-party library lz4 to version 1.9.4

striezel:
Files are taken from <https://github.com/lz4/lz4/tree/v1.9.4/lib>, the official Git repository of lz4.

For a list of changes between versions 1.9.1 and 1.9.4 see <https://github.com/lz4/lz4/releases>.


Zylann:
Thanks!

PR #534, opened 2023-07-15
Supress unused variable warning on VoxelBlockyModelMesh

DeeJayLSP:
On the current state, clang builds (required for macOS) will emit an unused variable warning.

`tangents_empty` isn't used outside of editor builds anyway.

This might make macOS checks (which use werror) work.


Zylann:
Thanks!

Issue #536, opened 2023-07-20
Allow specifying renering layer in VoxelInstanceLibraryMultiMeshItem


m4nu3lf:
**Is your feature request related to a problem? Please describe.**
I need to render some instances on a different layer than layer one because they should not be visible from a camera while the terrain must be visible.

**Describe the solution you'd like**
As for collision layers, it's reasonable also to have rendering layers for the multimesh.

**Describe alternatives you've considered**
I couldn't find any alternative.

Zylann:
Probably needs to be exposed somewhat

PR #537, opened 2023-07-20
Add renering layer for VoxelLibraryMultiMeshItem

m4nu3lf:
Addresses #536


Zylann:
Looks good, perhaps it also needs to copy the render layer in the scene setup: https://github.com/Zylann/godot_voxel/blob/99338703542685c1f92518cc23343f26d2287eb1/terrain/instancing/voxel_instance_library_multimesh_item.cpp#L256

m4nu3lf:
Yes, there were a few missing pieces. I've updated it.

Zylann:
Thanks!

PR #538, opened 2023-08-02
add github actions android builds and linux export template

magbulogtong81:
this PR adds a build for android, including export templates and android editor for all architectures. it also adds linux exports templates for i686 and x86_64 architectures.
please note that I have added concurrency on all yml files. this will cancel already running github actions if there's a new commit, which prioritizes the new commit


.github/workflows/android_builds.yml
```diff
@@ -0,0 +1,201 @@
...
+      #     python --version
+      #     scons --version
+
+      # We should always be explicit with our flags usage here since it's gonna be sure to always set those flags
+      - name: Compilation
+        env:
+          SCONS_CACHE: ${{github.workspace}}/.scons_cache/
+        run: |
+          PATH=${GITHUB_WORKSPACE}/x86_64-godot-linux-gnu_sdk-buildroot/bin:$PATH
+          scons -j2 verbose=yes warnings=extra platform=android target=${{matrix.target}} arch=${{matrix.arch}}
```
aaronfranke, line 178: The arch names used in Godot's build system are `"x86_32", "x86_64", "arm32", "arm64"` with some aliases to help users pick the correct name. You are currently using `"x86", "x86_64", "armv7", "arm64v8"` in this pull request. Ideally, the names the CI uses should be the same as what Godot uses, since these are the names the final executables will have. So please change this.

.github/workflows/mono.yml
```diff
@@ -10,6 +10,11 @@ env:
   GODOT_BASE_BRANCH: "4.1"
   SCONS_CACHE_LIMIT: 4096
 
+
+concurrency:
```
aaronfranke, line 14: Inconsistent empty line amount.

Zylann:
`android_builds.yml` could be just named `android.yml`, so far each file is named that way.

With artifacts for all these targets, this is going to increase space usage significantly, but we'll see if it causes problems. If it does I'll probably remove builds that are less used.

Why did you separate `linux_exports.yml` from `linux.yml`? It should be possible to have them in the same file by adding a matrix entry (although only 4.1, since that file also does 4.2 dev and 4.0 for compatibility checks).

The concurrency option is a nice addition :)

magbulogtong81:
I don't know what happened, git doesn't seem to display the changes I've made on android.yml correctly

Zylann:
> I've noticed that you have made different matrix for single and double precision builds for linux. so I've seperated it since I'm not sure if it's needed for linux export templates as well

Double-precision builds are in the same matrix.

> I don't know what happened, git doesn't seem to display the changes I've made on android.yml correctly

Usually when making PRs I amend the last commit and use `git push origin mybranch --force` to replace the commit in the Github repo. That usually updates the PR.
But if you have made several commits over it since, I don't know if it works. Maybe you could try squashing your commits: https://docs.godotengine.org/en/stable/contributing/workflow/pr_workflow.html#the-interactive-rebase

.github/workflows/android.yml
```diff
@@ -0,0 +1,202 @@
...
+
+jobs:
+  android-template:
+    runs-on: "ubuntu-20.04"
+    name: ${{matrix.name}}
+
+    strategy:
+      matrix:
+        include:
+        ####TEPMPLATES####
```
RomanRobot, line 26:
```suggestion
        ####TEMPLATES####
```

magbulogtong81:
closing this PR cause this doesn't seem to be needed for main repo

PR #548, opened 2023-09-23
Add option to disable culling neighbors for VoxelBlockyModel

### LINKED ISSUE #546: Allow disabling culling for block meshes
In Minecraft, certain blocks such as leaves do not cull their faces when placed next to each other. This gives the leaves a fuller appearance. It would be nice to be able to disable culling for specific block meshes to get this appearance. I've tried using a custom mesh but that didn't seem to work. I'm still new to this plugin

Minecraft (no culling):

Godot (culling):
Screenshot_20230918_165636

spazzylemons:
Adds an property named `culls_neighbors` to `VoxelBlockyModel`. When enabled (the default), the behavior is unchanged. When disabled, all neighboring sides are rendered. This resolves #546.

With `culls_neighbors` enabled:

culls_neighbors_enabled

With `culls_neighbors` disabled:

culls_neighbors_disabled


Zylann:
Looks good, thanks!

Issue #555, opened 2023-09-29
Nested voxel modifiers


sjkillen:
**Is your feature request related to a problem? Please describe.**

I have a post-import script that converts mesh instances into voxel modifier meshes. The modifiers do not work unless they are immediate children of a voxel terrain. If an imported scene may contain multiple instances, they cannot be made scene root.

**Describe the solution you'd like**
For modifiers to work if they are nested children of Node3D nodes

**Workaround**
Collapse the parent structure of the modifiers after import

sjkillen:
For `NOTIFICATION_LOCAL_TRANSFORM_CHANGED`
Here https://docs.godotengine.org/en/stable/classes/class_node3d.html
> This is not received when the transform of a parent node is changed.

If we have nodes under the terrain update their children when their local transform changes, we might be able to get around subscribing to global transform changes

Zylann:
Something I find really annoying in this situation, is that the hierarchy seems useless here, it's just in the way and requires the engine to do extra work. What makes it required is the import workflow and the way Godot works I guess? Though I'm not fully understanding why your scripts can't also automate the flattening of hierarchy, so that everything gets setup optimally (Think of it like CollisionShapes of a RigidBody).

Nested modifiers seems like a natural thing to support, but didn't have the need for it, and the way Godot propagates events put me off the moment I had the suggestion in mind. In fact I'm not using modifiers much at the moment, I originally added them because it seemed cool to have, as I saw another voxel engine have something similar. But I moved on other things and didnt take time to update them.

sjkillen:
I can see it also being useful for parent/child transforms, but that's also achievable manually.

As far as I know, an EditorScenePostImport must return a scene which becomes a packed scene resource which, as far as I know, can only have a single root. So it's impossible to flatten the structure from an EditorScenePostImport
It might be possible from an editor import plugin, but that's less convenient.

In my mind, it wouldn't be ideal to merge meshes into a single mesh SDF as each one has limited bake precision -but that assumption comes from my limited understanding of their implementation.

If you don't think it would be too non-trivial, I might have a shot at implementing it.

Zylann:
Couldn't you import your scene as a terrain?
Or when you instantiate these imported scenes, can't you just pick their children and put them as child of the terrain? These workarounds seem relatively easy, compared to adding an entire support for nesting just to workaround Godot's import workflow in the first place.

But if you really want this instead, you could try implementing it, as I don't plan to work on it at the moment.

[...]

PR #558, opened 2023-09-30
[WIP] Enable nested VoxelModifiers

sjkillen:
Regarding #555

I was unsure of what to set the transform to initially when the mesh is parented as global transforms aren't accessible. It seems that NOTIFICATION_TRANSFORM_CHANGED is called after the parenting notification.

This implementation appears to work with both the mesh and sphere modifiers. Nested modifiers will react to parent transformations

### DIFF
--- a/modifiers/godot/voxel_modifier_gd.cpp
+++ b/modifiers/godot/voxel_modifier_gd.cpp
@@ -6,9 +6,7 @@
 
 namespace zylann::voxel::gd {
 
-VoxelModifier::VoxelModifier() {
-	set_notify_local_transform(true);
-}
+VoxelModifier::VoxelModifier() { }
 
 zylann::voxel::VoxelModifier *VoxelModifier::create(zylann::voxel::VoxelModifierStack &modifiers, uint32_t id) {
 	ZN_PRINT_ERROR("Not implemented");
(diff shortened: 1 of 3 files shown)


sjkillen:
It doesn't seem to work right if some nodes are saved to a separate scene, I suspect this is because of how NOTIFICATION_PARENTED works and that it is called before a node is attached to the full scene tree

Zylann:
Are modifiers getting updated when the terrain is moved? They should not (and that's one big reason I didn't do this)

modifiers/godot/voxel_modifier_gd.cpp
```diff
@@ -115,16 +85,55 @@ void VoxelModifier::_notification(int p_what) {
...
+
+					if (modifier->is_sdf()) {
+						zylann::voxel::VoxelModifierSdf *sdf_modifier =
+								static_cast<zylann::voxel::VoxelModifierSdf *>(modifier);
+						sdf_modifier->set_operation(to_op(_operation));
+						sdf_modifier->set_smoothness(_smoothness);
+					}
+					// This transform is wrong, but will be updated with NOTIFICATION_TRANSFORM_CHANGED
+					// After node enters the tree
+					modifier->set_transform(get_transform());
```
Zylann, line 119: In case the modifier already is registered in the terrain, then you can get a valid transform, because modifier transforms are relative to the terrain. It doesn't matter whether they are in the tree or not. And once that's done, it should not have to be set again, until the modifier or any of its parents below the terrain are changed.
So I'm not sure you should set a wrong transform here...

modifiers/godot/voxel_modifier_gd.cpp
```diff
@@ -115,16 +85,55 @@ void VoxelModifier::_notification(int p_what) {
...
+			update_configuration_warnings();
 			if (_volume != nullptr && is_inside_tree()) {
 				VoxelData &voxel_data = _volume->get_storage();
 				VoxelModifierStack &modifiers = voxel_data.get_modifiers();
 				zylann::voxel::VoxelModifier *modifier = modifiers.get_modifier(_modifier_id);
 				ZN_ASSERT_RETURN(modifier != nullptr);
 
 				const AABB prev_aabb = modifier->get_aabb();
-				modifier->set_transform(get_transform());
+				Transform3D terrain_local = _volume->get_global_transform().affine_inverse() * get_global_transform();
```
Zylann, line 135: This is now more expensive even for direct children. At least maybe it should take a shortcut if direct child.

modifiers/godot/voxel_modifier_gd.cpp
```diff
@@ -115,16 +85,55 @@ void VoxelModifier::_notification(int p_what) {
...
+				// would be a breaking change as it returns null if node is toplevel
+				if (volume == nullptr) {
+					Node3D *grandparent = get_parent_node_3d();
+					volume = Object::cast_to<VoxelLodTerrain>(grandparent);
+					while (grandparent != nullptr && volume == nullptr) {				
+						grandparent = grandparent->get_parent_node_3d();
+						volume = Object::cast_to<VoxelLodTerrain>(grandparent);
+					}
+				}
+				_volume = volume;
```
Zylann, line 102: Parenting logic should not be in transform notifications. Other code might break unexpectedly if no transformations occur.

modifiers/godot/voxel_modifier_gd.cpp
```diff
@@ -115,16 +85,55 @@ void VoxelModifier::_notification(int p_what) {
...
+					modifier->set_transform(get_transform());
+					_modifier_id = id;
+					// TODO Optimize: on loading of a scene, this could be very bad for performance because there could be,
+					// a lot of modifiers on the map, but there is no distinction possible in Godot at the moment...
+					post_edit_modifier(*_volume, modifier->get_aabb());
+				}
+			}
 
-		case Node3D::NOTIFICATION_LOCAL_TRANSFORM_CHANGED: {
+			update_configuration_warnings();
```
Zylann, line 127: Why should configuration warnings update when the transform changes?

modifiers/godot/voxel_modifier_gd.cpp
```diff
@@ -115,16 +85,55 @@ void VoxelModifier::_notification(int p_what) {
 				_modifier_id = 0;
 			}
 		} break;
+		case Node3D::NOTIFICATION_TRANSFORM_CHANGED: {
```
Zylann, line 88: I believe this is called everytime *any* parent of the modifier is moved. If the terrain or any of its parents move, we actually *don't want* modifiers to get updated. Imagine a planet with a hundred craters made with modifiers. If the planet moves in space (as is the case in the Solar System demo), that would cause all of the craters to remesh every frame, even though they actually dont change locally. That's one reason I didnt implement nesting originally.

modifiers/godot/voxel_modifier_gd.cpp
```diff
@@ -115,24 +85,63 @@ void VoxelModifier::_notification(int p_what) {
...
 				VoxelData &voxel_data = _volume->get_storage();
 				VoxelModifierStack &modifiers = voxel_data.get_modifiers();
 				zylann::voxel::VoxelModifier *modifier = modifiers.get_modifier(_modifier_id);
 				ZN_ASSERT_RETURN(modifier != nullptr);
 
 				const AABB prev_aabb = modifier->get_aabb();
-				modifier->set_transform(get_transform());
+				Transform3D terrain_local = _volume->get_global_transform().affine_inverse() * get_global_transform();
+				// If the terrain moved, the modifiers do not need to be updated.
+				if (terrain_local == modifier->get_transform()) {
```
Zylann, line 137: I thought of this too, unfortunately at this point it already required computing a bunch of transforms, which can also suffer from floating point precision issues so the check might still fail

modifiers/godot/voxel_modifier_gd.cpp
sjkillen: It seems that the PARENTING notification is fired for packed scenes before they are attached to the main scene, so the volume is never found and the modifier is not initialized.
As a workaround, I moved the parenting code to the transform update. I suspect that a transform notification will always be fired after a parenting event because nodes do not have a global transform before they exist in the tree, however, I could be wrong about this. It appears to work in my tests.

modifiers/godot/voxel_modifier_gd.cpp
sjkillen: Ah, sorry. This was an artifact from copying the parenting code over to the transformation notification. Does it only need to be run if the modifier was added to the terrain?

modifiers/godot/voxel_modifier_gd.cpp
sjkillen: I added a commit that will check whether the transform that is local to the terrain has changed, however, because the computation of this transform relies on global transforms, there is some floating point error and it will sometimes update modifiers when the terrain is moved.
Would it be acceptable to measure some epsilon difference between transforms and only update if transforms differ by that amount? I'm not sure if there are helpers for this sort of comparison in Godot.

modifiers/godot/voxel_modifier_gd.cpp
sjkillen: I observed some floating point errors causing unnecessary updates. Do you think it would work to check whether the transforms differ by some small amount instead of checking for equality?

modifiers/godot/voxel_modifier_gd.cpp
Zylann: I would still not rely on the transform notification for parenting, at least not alone, because it's not what it's for... Alternatives are to either make a special getter function that attempts to get the volume each time it is found null, or to defer the whole logic to cases where the volume enters the scene tree, but that also has annoying drawbacks such as reparenting the terrain causing total remeshing for no reason...

modifiers/godot/voxel_modifier_gd.cpp
Zylann: It may run only when the conditions checked are affected, so currently, only when `_volume` changes.

modifiers/godot/voxel_modifier_gd.cpp
Zylann: I would rather prefer nothing to happen at all if possible... but it seems like supporting nesting requires extra complexity/workarounds I didnt want to add originally (especially if direct children have to pay the same cost). I don't know how else it can be solved. I'm also not aware of nodes in Godot that have transform-dependent logic getting register all the way to a specific Node3D parent like this.

modifiers/godot/voxel_modifier_gd.cpp
sjkillen: For the very last part, do you mean the updates should be batched initially somehow?

sjkillen:
A few more revisions,
In particular, I tried to make the code as close to the current for the case of immediate children.
I extracted the logic to locate the volume and added a function to compare transforms using an epsilon so that moving the terrain does not update nested modifiers.
I didn't know where to put this math code, so it's in the same file - there is also still a WARN message for each time the modifier is updated

sjkillen:
Perhaps since there is a bit of an additional cost for the immediate child case, it would make sense to give a warning suggesting immediate children be used?

modifiers/godot/voxel_modifier_gd.cpp
Zylann: What I mean is that the fact the terrain gets added to the scene tree may already trigger updates on its own, but if modifiers also trigger individual updates on top of that, it could mean unnecessary updates depending on the timing. Although I'm not certain that this really matters, it might already have been a problem before your PR.

PR #566, opened 2023-10-08
Make raycast based on model's collision_aabbs

Lry722:
The `RaycastPredicteBlocky` did not actually check for model's `collision_aabbs`. Now I've made it do that.

### DIFF
--- a/edition/voxel_tool_terrain.cpp
+++ b/edition/voxel_tool_terrain.cpp
@@ -59,6 +59,8 @@ Ref<VoxelRaycastResult> VoxelToolTerrain::raycast(
 		const VoxelData &data;
 		const VoxelBlockyLibraryBase::BakedData &baked_data;
 		const uint32_t collision_mask;
+		const Vector3 p_from;
+		const Vector3 p_to;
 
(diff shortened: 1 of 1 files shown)


edition/voxel_tool_terrain.cpp
```diff
@@ -110,7 +111,7 @@ Ref<VoxelRaycastResult> VoxelToolTerrain::raycast(
 		if (library_ref.is_null()) {
 			return res;
 		}
-		RaycastPredicateBlocky predicate{ _terrain->get_storage(), library_ref->get_baked_data(), p_collision_mask };
+		RaycastPredicateBlocky predicate{ _terrain->get_storage(), library_ref->get_baked_data(), p_collision_mask, p_pos, p_pos + p_dir * max_distance};
```
Zylann, line 114: The raycast takes place in terrain space, however you are sending global space coordinates. Use `local_pos` and `local_dir` instead.

Lry722: Fixed

Zylann:
Thanks!

PR #594, opened 2024-01-21
Add do_surface function for VoxelTool

Piratux:
This PR adds `do_surface` function for `VoxelTool`. It's meant to be analogous to Surface tool from Unreal Engine Voxel Plugin.

Small demo of `do_surface` in action: https://imgur.com/a/NaWBZzs


Zylann:
> This PR adds do_surface function for VoxelTool. It's meant to be analogous to Surface tool from Unreal Engine Voxel Plugin.

Can you explain what it actually does? The name could even need to be changed based on that, because "surface" is really vague.

edition/voxel_tool.cpp
```diff
@@ -324,6 +324,39 @@ void VoxelTool::smooth_sphere(Vector3 sphere_center, float sphere_radius, int bl
...
+	Ref<gd::VoxelBuffer> buffer;
+	buffer.instantiate();
+	buffer->create(voxel_box.size.x, voxel_box.size.y, voxel_box.size.z);
+
+	if (_channel == VoxelBufferInternal::CHANNEL_SDF) {
+		// Note, this only applies to SDF. It won't affect voxel texture data.
+		
+		copy(voxel_box.pos, buffer, (1 << VoxelBufferInternal::CHANNEL_SDF));
+
+		std::shared_ptr<VoxelBufferInternal> smooth_buffer = make_shared_instance<VoxelBufferInternal>();
```
Zylann, line 347: std::shared_ptr is not necessary, VoxelBufferInternal can be created on the stack

edition/voxel_tool.cpp
```diff
@@ -324,6 +324,39 @@ void VoxelTool::smooth_sphere(Vector3 sphere_center, float sphere_radius, int bl
...
+void VoxelTool::do_surface(Vector3 sphere_center, float sphere_radius, float strength) {
+	ZN_PROFILE_SCOPE();
+	ZN_ASSERT_RETURN(sphere_radius >= 0.01f);
+	ZN_ASSERT_RETURN(strength >= 0.0f && strength <= 1.0f);
+
+	const Box3i voxel_box = Box3i::from_min_max(
+			math::floor_to_int(sphere_center - Vector3(sphere_radius, sphere_radius, sphere_radius)),
+			math::ceil_to_int(sphere_center + Vector3(sphere_radius, sphere_radius, sphere_radius)));
+
+	// TODO Perhaps should implement `copy` and `paste` with `VoxelBufferInternal` so Godot object wrappers wouldn't be
```
Zylann, line 336: This; but can be done later, I've done that for copy, just havent done yet for paste

Piratux:
I agree that the name does not speak itself. I just chose word "surface" because that's what it's called in UE Voxel Plugin.

While there are similarities with `do_sphere` function (ignoring that it does min/max), the issue is that, `do_sphere` also modifies voxels outside the sphere. So if `GROW`/`SHRINK` were to be added, `do_sphere` (and other shapes) would need to only modify voxels inside the shape (if you agree with this solution, then I'm fine with that, and `do_surface` would not be needed anymore). Something to keep in mind with this though, is that these modes would only really be needed for sphere. So it may end up being waste of time having to implement it for other shapes.

And sorry for my messy implementation. I knew the code would not be ideal, just mainly wanted to get idea out, so it can be discussed how it can be improved, or if it's needed at all. I'll update code with the reviewed comments, if we decide that it's better to add `do_surface` function over additional `GROW`/`SHRINK` modes.

I haven't read how docs are generated, so I updated both .md and .xml file descriptions by hand. I'll keep in mind, that only a single one is needed to be filled in next time.

Zylann:
> do_sphere also modifies voxels outside the sphere

That's just an optimization. You actually still modify them in your code, just setting the same value back. This is a different operation so nothing prevents from keeping that optimization. 

> if GROW/SHRINK were to be added, do_sphere (and other shapes) would need to only modify voxels inside the shape

Why so? It's a different mode, therefore it can have a different code path when selecting voxels to process. Probably just needs some investigation to see how that can be done, as this area uses some template magic.
Also the reason I mention using a mode is not just for automatically supporting other shapes, but also because there are other optimizations already in place that we'd get for free, and potentially needs less code.

> And sorry for my messy implementation

There is redundancy and a few things I noted, but it's actually quite clear to understand. It's a distinct operation so it may be added in some form, just not decided yet which one will be best, I'd prefer a mode ideally.

> I haven't read how docs are generated, so I updated both .md and .xml file descriptions by hand. I'll keep in mind, that only a single one is needed to be filled in next time.

If you're curious, you can run `build.py` from this folder: https://github.com/Zylann/godot_voxel/tree/master/doc/tools
Although I prefer making dedicated commits for this because it will update everything, including last change dates and stuff

Piratux:
> That's just an optimization. You actually still modify them in your code, just setting the same value back. This is a different operation so nothing prevents from keeping that optimization.
What I mean here, is that currently each shape expects SDF value to be returned for each position (see https://github.com/Zylann/godot_voxel/blob/master/edition/funcs.h#L236). So in this case, function that calls the operator to get SDF value of shape, would need to call it only for voxels inside the shape (in this case sphere) or pass the current voxel value.

To get ahead of myself (for another PR I'd like to do next), while it may not seem ideal, passing current voxel value would allow implementing "falloff" feature, where shape appears "gradually" when applied in same location multiple times.

But yeah, I agree, additional modes, would probably be a preferable solution over `do_surface` suggested in this PR. I could attempt to come up with implementation for this, but chances are low I'd get it right.

Zylann:
> What I mean here, is that currently each shape expects SDF value to be returned for each position

Not necessarily, for texture painting I added a fastpath doing that exact optimization. That function call can be avoided in certain cases. The only issue is that I didnt give much thought on making it generic enough using templates, so as a result you will notice there is some code duplication going on.

Tbh it's possible that another way of implementing all operations would be preferred in the future, because I'm not 100% convinced with all the current edition code, notably considering a future SIMD path. For example if ISPC were to be added, it's possible that heavily templated code wouldn't map well, but on the other hand, I don't want to have to constantly repeat/re-invent ways of writing similar operations for every case, which is why I wrote that template framework.

Eventually, after adressing the name and several code issues that function could be merged if that's more convenient, but it's possible that it gets changed in a future rework.

edition/voxel_tool.cpp
Piratux: Removed the comment

Piratux:
Pushed changes in last commit:
- Renamed `do_surface` to `grow_sphere`.
- Removed manual documentation from `.md` about `grow_sphere`.
- Removed unnecessary second buffer for `grow_sphere` operation.
- Fixed mentioned things in review comments.

Zylann:
Some CI jobs are failing but that's just because your PR doesn't have a fix I added more recently in unrelated areas. I'll merge it and do some adjustments.

PR #598, opened 2024-02-04
Create universal macOS build with Intel/M-series chips support.

marknefedov:


.github/workflows/macos.yml
```diff
@@ -79,7 +79,9 @@ jobs:
         env:
           SCONS_CACHE: ${{github.workspace}}/.scons_cache/
         run: |
-          scons -j2 verbose=yes warnings=extra werror=yes platform=macos tests=no target=${{matrix.target}} dev_build=no
+          scons -j$(sysctl -n hw.ncpu) arch=x86_64 verbose=yes warnings=extra werror=yes platform=macos tests=no target=${{matrix.target}} dev_build=no
```
Zylann, line 82: It isn't necessary to specify `-j` if you want all cores to be used, SCons will auto-detect them

Zylann:
Looks like the cache broke, the job took 1h30 (and likely always will). There is a bit more to do in order to make it work again. Two builds are done in serial here, but there is only one cache. Maybe 2 directories are needed, and 2 caches as well with the right name so they don't conflict. 
Or maybe something simpler would be to just add an `arch` parameter to the matrix, and have another job invoking the first one, which would get both artifacts to generate the third universal one?

marknefedov:
Maybe I'm not sure actually, at least https://docs.godotengine.org/en/stable/contributing/development/compiling/compiling_for_macos.html#building-export-templates show how to make universal binary for templates. Probably need to compose an .app archive.

Zylann:
I'll assume the universal build is what Godot wants, it might just be a file extension thing. The fact the doc mentions a `.app` and the export templates window expects `.zip` is confusing.

I'm still considering universal builds so I'll likely integrate this to the CI at some point.

PR #607, opened 2024-02-29
fix: fix broken builds because of godot physics files moved to another directory

Lunatix89:
The title says it all, I've had problems building from source because the godot devs moved all physics related files to `scenes/3d/physics`.

### DIFF
--- a/util/godot/classes/rigid_body_3d.h
+++ b/util/godot/classes/rigid_body_3d.h
@@ -2,7 +2,14 @@
 #define ZN_GODOT_RIGID_BODY_3D_H
 
 #if defined(ZN_GODOT)
+#include <core/version.h>
+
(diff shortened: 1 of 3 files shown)


util/godot/classes/rigid_body_3d.h
```diff
@@ -2,7 +2,14 @@
 #define ZN_GODOT_RIGID_BODY_3D_H
 
 #if defined(ZN_GODOT)
-#include <scene/3d/physics_body_3d.h>
+#include <core/version.h>
+
+#if VERSION_MAJOR == 4 && VERSION_MINOR <= 2
+#include <scene/3d/rigid_body_3d.h>
```
Zylann, line 8: Should be `#include <scene/3d/physics_body_3d.h>`

Lunatix89: Sorry, missed that one

PR #611, opened 2024-03-27
SQLite: Update to 3.45.2

Chubercik:
For a full changelog see:

- https://sqlite.org/changes.html

(previously used version - 3.42.0)

---

Files are taken from SQLite's source code amalgamation for the version 3.45.2:

- https://sqlite.org/2024/sqlite-amalgamation-3450200.zip (2.61 MiB)

---

Additionally changed `thirdparty/sqlite/version.md` to `thirdparty/sqlite/VERSION.txt` (there is no need to store it as a Markdown file).


Zylann:
Thanks!

PR #616, opened 2024-04-02
Small grow_sphere adjustments

Piratux:
Changes:
- Fixed `grow_sphere` not using `sdf_scale` in its calculations.
- Increased allowed strength parameter range for `grow_sphere` from [0, 1] to [0, 10].

### DIFF
--- a/edition/voxel_tool.cpp
+++ b/edition/voxel_tool.cpp
@@ -319,7 +319,7 @@ void VoxelTool::grow_sphere(Vector3 sphere_center, float sphere_radius, float st
 	// see: https://github.com/Zylann/godot_voxel/pull/594
 	ZN_PROFILE_SCOPE();
 	ZN_ASSERT_RETURN(sphere_radius >= 0.01f);
-	ZN_ASSERT_RETURN(strength >= 0.0f && strength <= 1.0f);
+	ZN_ASSERT_RETURN(strength >= 0.0f && strength <= 10.0f);
 
 	const Box3i voxel_box = Box3i::from_min_max(
 			math::floor_to_int(sphere_center - Vector3(sphere_radius, sphere_radius, sphere_radius)),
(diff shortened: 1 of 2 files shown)


edition/voxel_tool.cpp
```diff
@@ -319,7 +319,7 @@ void VoxelTool::grow_sphere(Vector3 sphere_center, float sphere_radius, float st
 	// see: https://github.com/Zylann/godot_voxel/pull/594
 	ZN_PROFILE_SCOPE();
 	ZN_ASSERT_RETURN(sphere_radius >= 0.01f);
-	ZN_ASSERT_RETURN(strength >= 0.0f && strength <= 1.0f);
+	ZN_ASSERT_RETURN(strength >= 0.0f && strength <= 10.0f);
```
Zylann, line 322: Considering the code was assuming exposed SDF range was normalized (-1..1), and recently changed to be proper distances, it means APIs now need a *500 factor to have the same behavior (since the internal scale to normalize is 0.002 in 16-bit). So perhaps this needs to be capped to 500 instead of 10?
10 would be the right cap for 8-bit, but 500 would work for both since either way the result is capped when encoded.
Unless you found 10 is enough of a maximum by leaving `sdf_scale` to 1?

Piratux: Honestly, I'm not sure if this needs capping in the first place.

The reason I've changed it to range [0, 10], is because from my experiments this is the only range I've used that fit well with slow/fast erosion preferences.

So previously, if formula looked like this (with `sdf_scale` being multiplied):
`sdf_scale = 0.002`
`sdf_scale * strength`
`strength` parameter should have been 500 times bigger.

But now that it looks like this by default:
`sdf_scale = 1`
`sdf_scale * strength`
`strength` fits well being in range [0.1, 10] or [0, 10].

Zylann: `sdf_scale` was 0.002 before and now is 1.0, so if we keep the same ratios indeed `strength` (which was 0..1 before) would have to be multiplied by 500 (therefore becoming 0..500). I don't see the logic in the range becoming 0..10, or it's arbitrary. But if we forget about whatever it was before maybe it doesn't matter in the end?
So capped? Not capped?

Piratux: I've forgot, that I had `sdf_scale` set to `10` in my code.
So having tested new and old behavior, I've noticed that old behavior with `strength=0.1`, matches exactly new behavior with `strength=10`.
So it's `*100` instead of `*500` (which I'd guess is due to `SDF_FAR_INSIDE`/`SDF_FAR_OUTSIDE` values).

But I'll just remove the strength limits, considering it's based on preferences.

PR #627, opened 2024-04-17
Fix crash when saving an empty VoxelBlockyType

ArchLinus:
The editor will crash if you try to save a scene with an empty VoxelBlockyType.

This PR fixes that by adding a null check.


Zylann:
Thanks!

PR #629, opened 2024-04-29
fix: editor_plugin.h was moved to editor/plugins

Lunatix89:
Looks like `editor/editor_plugin.h` was moved to `editor/plugins/` so the build fails with the latest master branch (see here: https://github.com/godotengine/godot/pull/90975).


Lunatix89:
There is also another reference in `modules/voxel/util/godot/classes/editor_interface.h` but I'm not sure whether this needs to be fixed as well or not.
I'm also not sure whether it would be better to add a switch like "#if VERSION_MAJOR == 4 && VERSION_MINOR >= 3` because it would otherwise fail builds in older releases.

Zylann:
Should be fixed in 2c625a6ba29488ad12ca31254bc284d45eb70eb4

PR #672, opened 2024-07-06
Fix fuzzer

Piratux:
Fixed fuzzer issues:
- `System.IO.IOException: No space left on device`: fixed by adding `Free disk space` step before compilation step.
- Then there was an issue of step trying to fetch 4.2 version of `qarmin/Qarminer` repository, but only 4.1 exists there. I made an [issue](https://github.com/qarmin/Qarminer/issues/32) about it, but the author said that 4.1 version should work with Godot version up to 4.3+. So I hardcoded it to 4.1 for now. This may need to be updated later.

Fuzzer working and catching Godot crashes can be seen here:
<https://github.com/Piratux/godot_voxel/actions/runs/9821448650>

### DIFF
--- a/.github/workflows/fuzzer.yml
+++ b/.github/workflows/fuzzer.yml
@@ -117,6 +117,8 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: qarmin/Qarminer
+          # At the time of writing, qarmin did not put out later than 4.1 versions, but claims that it works with Godot versions up to 4.3+
+          # See https://github.com/qarmin/Qarminer/issues/32
           ref: 4.1
           path: fuzzer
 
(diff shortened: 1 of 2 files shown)


.github/workflows/fuzzer.yml
```diff
@@ -107,7 +117,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: qarmin/Qarminer
-          ref: ${{ env.GODOT_BASE_BRANCH }}
+          ref: 4.1
```
Zylann, line 120: Can you add a comment explaining why this is the chosen version?

Piratux: Done

PR #698, opened 2024-09-05
modifiers: support nesting

NoTalkOnlyProx:
This PR improves/adds support for nested modifier structures in the voxel lod terrain tree.

These changes are not of particular importance to me (other than the ability to filter out terrain translation). I just happened to be in the area and noticed your TODO comment, and I thought this might be a good place to cut my teeth.

### DIFF
--- a/modifiers/godot/voxel_modifier_gd.cpp
+++ b/modifiers/godot/voxel_modifier_gd.cpp
@@ -11,7 +11,7 @@
 namespace zylann::voxel::godot {
 
 VoxelModifier::VoxelModifier() {
-	set_notify_local_transform(true);
+	set_notify_transform(true);
 }
 
 zylann::voxel::VoxelModifier *VoxelModifier::create(zylann::voxel::VoxelModifierStack &modifiers, uint32_t id) {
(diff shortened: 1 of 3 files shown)


NoTalkOnlyProx:
oops. Sorry. I just noticed #555 and #558.

Time to review the other approach and make sure I didn't miss anything, I guess!

NoTalkOnlyProx:
One thing I notice off the bat is that I did not implement this in my version:

```
void VoxelModifier::mark_as_immediate_child(bool v) {
		} break;
	_is_immediate_child = v;
	if (_is_immediate_child) {
		set_notify_local_transform(true);
		set_notify_transform(false);
	} else {
		set_notify_local_transform(false);
		set_notify_transform(true);
	}
}
```

I considered doing something essentially identical, but I decided that the performance gains (which if I am not mistaken aren't much?) were not worth cluttering the solution.

Please let me know if you disagree with that judgement, though!

NoTalkOnlyProx:
That also appears to be the only substantive difference between our two PRs

Zylann:
I'm still not happy with the workarounds that Godot requires to make this work. Just like the other PR, I'm not sure about merging this.

PR #708, opened 2024-10-06
Various MSVC compilation fixes

Repiteo:
Assortment of tweaks to support the arguments `warnings=extra werror=yes dev_build=yes d3d12=yes` in an MSVC context

### DIFF
--- a/generators/graph/node_type_db.cpp
+++ b/generators/graph/node_type_db.cpp
@@ -244,7 +244,7 @@ VoxelGraphFunction::Port make_port_from_io_node(const ProgramGraph::Node &node, 
 }
 
 bool is_node_matching_port(const ProgramGraph::Node &node, const VoxelGraphFunction::Port &port) {
-	if (node.type_id != port.type) {
+	if (node.type_id != static_cast<uint32_t>(port.type)) {
 		return false;
 	}
 
--- a/meshers/transvoxel/transvoxel.cpp
+++ b/meshers/transvoxel/transvoxel.cpp
@@ -7,6 +7,10 @@
 #include "../../util/profiling.h"
 #include "transvoxel_tables.cpp"
 
+#if defined(_MSC_VER)
+#pragma warning(disable : 4701) // Potentially uninitialized local variable used.
+#endif
+
 // #define VOXEL_TRANSVOXEL_REUSE_VERTEX_ON_COINCIDENT_CASES
 
 namespace zylann::voxel::transvoxel {
--- a/streams/sqlite/connection.cpp
+++ b/streams/sqlite/connection.cpp
@@ -417,6 +417,7 @@ bool Connection::save_block(const BlockLocation loc, const Span<const uint8_t> b
 			update_block_statement = _update_instance_block_statement;
 			break;
 		default:
+			update_block_statement = nullptr;
 			CRASH_NOW();
 	}
 
@@ -471,6 +472,7 @@ VoxelStream::ResultCode Connection::load_block(
 			get_block_statement = _get_instance_block_statement;
 			break;
 		default:
+			get_block_statement = nullptr;
 			CRASH_NOW();
 	}
 
--- a/terrain/instancing/voxel_instance_generator.cpp
+++ b/terrain/instancing/voxel_instance_generator.cpp
@@ -9,6 +9,10 @@
 #include "../../util/math/triangle.h"
 #include "../../util/profiling.h"
 
+#if defined(_MSC_VER)
+#pragma warning(disable : 4701) // Potentially uninitialized local variable used.
+#endif
+
 namespace zylann::voxel {
 
 namespace {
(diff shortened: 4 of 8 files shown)


generators/graph/node_type_db.cpp
```diff
@@ -244,7 +244,7 @@ VoxelGraphFunction::Port make_port_from_io_node(const ProgramGraph::Node &node,
 }
 
 bool is_node_matching_port(const ProgramGraph::Node &node, const VoxelGraphFunction::Port &port) {
-	if (node.type_id != port.type) {
+	if (node.type_id != (uint32_t)port.type) {
```
Zylann, line 247: Use `static_cast`

meshers/transvoxel/transvoxel.cpp
```diff
@@ -1237,7 +1237,7 @@ void build_transition_mesh(
 									cell_textures.weights[index_vertex_a];
 							const FixedArray<uint8_t, MAX_TEXTURE_BLENDS> weights1 =
 									cell_textures.weights[index_vertex_b];
-							FixedArray<uint8_t, MAX_TEXTURE_BLENDS> weights;
+							FixedArray<uint8_t, MAX_TEXTURE_BLENDS> weights = {};
```
Zylann, line 1240: What is this doing? Initializing earlier is pointless because the following for loop is doing it.

streams/sqlite/connection.cpp
```diff
@@ -408,7 +408,7 @@ bool Connection::save_block(const BlockLocation loc, const Span<const uint8_t> b
 
 	sqlite3 *db = _db;
 
-	sqlite3_stmt *update_block_statement;
+	sqlite3_stmt *update_block_statement = nullptr;
```
Zylann, line 411: Initialization happens in the switch statement or exits it (via crash), so why would it need to be initialized here too? Doesn't this incur a cost? Times where I had to fix this kind of warnings, I assigned a default value in the `default` case

terrain/instancing/voxel_instance_generator.cpp
```diff
@@ -507,7 +507,7 @@ void VoxelInstanceGenerator::generate_transforms(
 		bool surface_normal_is_normalized = false;
 		bool sphere_up_is_computed = false;
 		bool sphere_distance_is_computed = false;
-		float sphere_distance;
+		float sphere_distance = 0.f;
```
Zylann, line 510: `sphere_distance` is not always used, but cases where it is used have it initialized. Same story here: if it gets initialized here, won't that incur a cost when configurations such that it's not actually used? It's quite annoying

thirdparty/fast_noise_2/src/FastSIMD/FastSIMD.cpp
```diff
@@ -215,7 +215,7 @@ CLASS_T* SIMDLevelSelector( FastSIMD::eLevel maxSIMDLevel, FastSIMD::MemoryAlloc
             return nullptr;
         }
 
-        return SIMDLevelSelector<CLASS_T, FastSIMD::SIMDTypeList::GetNextCompiledAfter<SIMD_LEVEL>>( maxSIMDLevel, allocator );
```
Zylann, line 218: Don't modify FastNoise2. It's a thirdparty library. Warnings are turned off on it for that reason. If you still can't compile it, you should change compilation flags specifically for its files, or refer to the library's author maybe?

util/string/expression_parser.cpp
```diff
@@ -88,7 +88,7 @@ bool get_number_token(const std::string_view text, unsigned int &pos, Token &out
 	}
 
 	// Decimal part
-	double f;
+	double f = 0;
```
Zylann, line 91: If this default value is necessary, that means the logic following it found an incorrect format... though that's not really handled at the moment. I wonder if compilers generating this warning are still able to optimize this away, because the fact they emit that warning in the first place sounds like they can't...

thirdparty/fast_noise_2/src/FastSIMD/FastSIMD.cpp
Repiteo: Actually, looks like theyr'e *not* turned off. At least, not for MSVC; only a single warning is disabled for that thirdparty library. I'll just make it disable warnings universally instead

meshers/transvoxel/transvoxel.cpp
Repiteo: Seems like MSVC being dumb
```
D:\Godot\Github\godot\modules\voxel\meshers\transvoxel\transvoxel.cpp(1246) : error C2220: the following warning is treated as an error
D:\Godot\Github\godot\modules\voxel\meshers\transvoxel\transvoxel.cpp(1246) : warning C4701: potentially uninitialized local variable 'weights' used
```
Looks like there's already some contingency in place for `fixed_array.h` (lines 56-61), so this might be in a similar boat?

Repiteo: For this and the other 3 that didn't have any real "clean" resolution, I added an MSVC define to the top for handling these cases. Couldn't surround the affected areas with push/pop wrappers for some reason, but they're not header files to begin with so it's an acceptable fallback imo

PR #728, opened 2025-02-06
Fix save_generator_output not working in VoxelLodTerrain.

SiriusRP:
cache_generated_blocks is passed to LoadBlockDataTask where it is used to check if a block should be generated and saved if not found by the stream. However, it default constructs to false and is never written to. Meaning this check will always fail.

https://github.com/Zylann/godot_voxel/blob/f79a29f418554049d91a85e7dfdd8d415dea557e/streams/load_block_data_task.cpp#L67-L97

Setting this to construct as true fixes the issue and better aligns with how VoxelTerrain is written. But if there is a better option for fixing this I'm happy to change it.

### DIFF
--- a/terrain/variable_lod/voxel_lod_terrain_update_data.h
+++ b/terrain/variable_lod/voxel_lod_terrain_update_data.h
@@ -72,7 +72,7 @@ struct VoxelLodTerrainUpdateData {
 		// If true, try to generate blocks and store them in the data map before posting mesh requests.
 		// If false, meshing will generate non-edited voxels on the fly instead.
 		// Not really exposed for now, will wait for it to be really needed. It might never be.
-		bool cache_generated_blocks = false;
+		bool cache_generated_blocks = true;
 		bool collision_enabled = true;
 		bool detail_textures_use_gpu = false;
 		bool generator_use_gpu = false;


Zylann:
It is not written to because I wanted to investigate and test before introducing it in this terrain type, meaning that feature is indeed not officially supported here at the moment. I don't remember right now all the details that this implies, but I know it might have some edge cases.
Another reason is that for this terrain type, I haven't needed that caching yet, though I can see a use to speed up frequent voxel lookups. Also, it consumes a LOT more memory, and saving to stream can reduce the effectiveness of detail rendering.
There might be considerations to take with full loading mode, which, while not using that flag currently, could also use caching somehow.

Note that this boolean is *not* directly about saving, it is about caching. Saving indeed only occurs if you *also* enable that option on the stream. There could be ways to save without caching, which would have to happen in the meshing task, but that seems weird.

If you make it default to true, you are also not fixing the fact that nothing sets it. You'll just make the terrain cache everything all the time and use more memory (assuming it works), which some projects might not want.

Basically, this option isn't really ready to be exposed or changed yet.

SiriusRP:
Do you have any suggestions for this should be implemented?

I have a custom generator that reads GIS data into VoxelLodTerrain. Currently it is very slow and I was hoping to speed up this process by writing the generated blocks to disk and then streaming them on subsequent loads.

Zylann:
> Do you have any suggestions for this should be implemented?

Not currently. My free time is currently blocked by something else so I can't work on my projects much.

> I have a custom generator that reads GIS data into VoxelLodTerrain. Currently it is very slow and I was hoping to speed up this process by writing the generated blocks to disk and then streaming them on subsequent loads.

If you want to save this into a stream, you could consider doing this process ahead of time maybe. You could maybe even consider implementing a custom stream, or figuring how to optimize what is slow. If this is slow because you're doing web requests though, you might be taking the wrong approach IMO, since you would end up with all task threads blocked *waiting for requests, not doing actual work*, the system isn't optimized for that.
You're also welcome to test your current change with your project more in depth, even if I don't expose that yet.
Unfortunately for the same reasons stated above I can't offer deeper help on subjects like this at the moment,

Zylann:
I pushed an experimental property that exposes the variable in bfc72a9f3fc7a5db978641faa60bc86d7fbed707

PR #735, opened 2025-03-07
Fix for GPU Generation Hitting Rid Allocation Limit

mschwarz1:
I was running into an issue where after a couple of minutes of flying around with gpu generation on where I would start hitting the rid allocation limit. I took a look at the code and noticed the generator_uniform_set variable was being created but never cleared as far as I could tell. So I switched it to a member variable like _generator_pipeline_rid and freed it along with that and it seems to fix the problem. 

I didn't see any resource allocation issues after ~15 minutes of flying around with this change where as before I was hitting the issue after ~3 minutes.

### DIFF
--- a/generators/generate_block_gpu_task.cpp
+++ b/generators/generate_block_gpu_task.cpp
@@ -151,8 +151,8 @@ void GenerateBlockGPUTask::prepare(GPUTaskContext &ctx) {
 		// Note, this internally locks RenderingDeviceVulkan's class mutex. Which means it could perhaps be used outside
 		// of the compute list (which already locks the class mutex until it ends). Thankfully, it uses a recursive
 		// Mutex (instead of BinaryMutex)
-		_generator_uniform_set =
-				zylann::godot::uniform_set_create(rd, generator_uniforms, generator_shader_rid, 0);
+		const RID generator_uniform_set = zylann::godot::uniform_set_create(rd, generator_uniforms, generator_shader_rid, 0);
+		_generator_uniform_sets.push_back(generator_uniform_set);
 
 		{
 			ZN_PROFILE_SCOPE_NAMED("compute_list_bind_compute_pipeline");
@@ -160,7 +160,7 @@ void GenerateBlockGPUTask::prepare(GPUTaskContext &ctx) {
 		}
 		{
 			ZN_PROFILE_SCOPE_NAMED("compute_list_bind_uniform_set");
-			rd.compute_list_bind_uniform_set(compute_list_id, _generator_uniform_set, 0);
+			rd.compute_list_bind_uniform_set(compute_list_id, generator_uniform_set, 0);
 		}
 
 		const Box3i &box = boxes_to_generate[box_index];
@@ -464,7 +464,10 @@ void GenerateBlockGPUTask::collect(GPUTaskContext &ctx) {
 	}
 
 	zylann::godot::free_rendering_device_rid(rd, _generator_pipeline_rid);
-	zylann::godot::free_rendering_device_rid(rd, _generator_uniform_set);
+
+	for (RID rid : _generator_uniform_sets) {
+		zylann::godot::free_rendering_device_rid(rd, rid);
+	}
 	
 	for (RID rid : _modifier_pipelines) {
 		zylann::godot::free_rendering_device_rid(rd, rid);
--- a/generators/generate_block_gpu_task.cpp
+++ b/generators/generate_block_gpu_task.cpp
@@ -151,7 +151,7 @@ void GenerateBlockGPUTask::prepare(GPUTaskContext &ctx) {
 		// Note, this internally locks RenderingDeviceVulkan's class mutex. Which means it could perhaps be used outside
 		// of the compute list (which already locks the class mutex until it ends). Thankfully, it uses a recursive
 		// Mutex (instead of BinaryMutex)
-		const RID generator_uniform_set =
+		_generator_uniform_set =
 				zylann::godot::uniform_set_create(rd, generator_uniforms, generator_shader_rid, 0);
 
 		{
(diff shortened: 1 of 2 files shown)


Zylann:
Back when I first tested GPU stuff with Godot I started the dev project for making shaders in the module's repo, and freeing uniform sets was commented out. I don't remember why, but I notice that I consistently never freed uniform sets. Forgot the reason as well.
But according to docs it should be freed, so there are more places to do this in.

generators/generate_block_gpu_task.cpp
```diff
@@ -151,7 +151,7 @@ void GenerateBlockGPUTask::prepare(GPUTaskContext &ctx) {
 		// Note, this internally locks RenderingDeviceVulkan's class mutex. Which means it could perhaps be used outside
 		// of the compute list (which already locks the class mutex until it ends). Thankfully, it uses a recursive
 		// Mutex (instead of BinaryMutex)
-		const RID generator_uniform_set =
+		_generator_uniform_set =
```
Zylann, line 154: This is in a for loop. You will still leak RIDs in different setups.

mschwarz1: Ahh yeah tunnel vision. I'll have to fix that. I'm surprised the change is making a much of a difference at all given that added context.

mschwarz1:
> Thanks for spotting it!
> 
> Back when I first tested GPU stuff with Godot when they just came out, I started the dev project for making shaders in the module's repo, and freeing uniform sets was commented out. I don't remember why, but I notice that I consistently never freed uniform sets. Forgot the reason as well. But according to docs it should be freed, so there are more places to fix.

It's funny you mention that because I have another script where I was also trying to free the uniform set because I was following the documentation but it resulted in saying I was freeing something invalid which I wasn't expecting. In this case it's not complaining about anything so I assume it's a valid thing to clear . Point being I'm not really sure what the deal is with uniform sets either.

generators/generate_block_gpu_task.cpp
Zylann: Multi-box situations occur in particular cases (notably edited terrain) that you might not have encountered.

mschwarz1: That makes sense. The other places using uniform_set_create are in the render_detail_texture_gpu_task which I also wouldn't have seen. So I'll try turning that on and doing some edits in the next pass.

Zylann:
I asked devs, and apparently uniform sets are supposed to be "freed by their dependencies", which hopefully means if one of their contents is freed, then the uniform set is freed. This is still unclear to me because none of this is documented...

Considering uniform sets are then supposed to be automatically freed...  I'm also told the issue could be something else. 
Are you able to provide a minimal test project that reproduce this issue?

mschwarz1:
[minimaltest.zip](https://github.com/user-attachments/files/19134612/minimaltest.zip)

Attached is a minimal reproduction project. All you need to do is play the one scene and wait 2 minutes or so. Things to note. I'm running a pretty much 4.4 double precision build of godot. (I really hope this isn't another random double build specific behavior being weird.) I don't notice anything because of the double precision but if you start seeing the mesh looking strange it might just be because the camera is getting far out. You should see the issues around 21km though so I don't think you'd run into anything super noticeable until you hit the allocation issue and then you'll really notice some weirdness along with the printed errors.

mschwarz1:
An aside I tried out having detail normal maps enabled with the gpu generation on my other project where I ran into this originally after making the for loop update and I didn't seem to hit any issues. Not sure why. Maybe I didn't let it go long enough or those uniform_sets are being cleaned up automatically?

Zylann:
So far my tests just show that without my branch, moving forwards periodically reaches this line: https://github.com/godotengine/godot/blob/7459a0361d7cea65e190232ae60c6fcc275c73f5/core/templates/rid_owner.h#L127
And every time, it is to allocate a RID for one of those uniform sets. `chunk_count` is bigger each time, so if I keep going it would eventually reach the limit.
But with my branch, that never happens (because I free the uniform sets manually).

Something new though:
One of the uniforms in the set changing each time. It is a storage buffer. However, I never free this storage buffer either, because I pool them. Which means that technically, none of the RIDs referenced by the uniform buffer get freed. And that's expected (some because they are pooled, others because they indeed dont need to change across tasks). Which means Godot's auto-freeing behavior will not trigger. And that we really need to free those uniform buffers manually then.

mschwarz1:
The pooling would explain why I didn't see anything with the normal maps assuming those uniform sets aren't part of a pooled resource. 

See my edits to the comment with my minimal reproduction project if you didn't already. I don't think you'll need the new version if you're just tracking the chunk_count directly though.

Zylann:
> The pooling would explain why I didn't see anything with the normal maps assuming those uniform sets aren't part of a pooled resource.

Not exactly. The pooling of *storage buffers* is what makes the *GPU chunk generation* not auto-free uniform sets. Because then Godot sees that all resources referenced by the uniform remain alive.
Normal maps might actually have the same issue, I think they use pooled buffers too. However, I think textures are not pooled though, which might be the reason.
But you have to also enable GPU usage for them in terrain properties (neither of your test projects have this).

mschwarz1:
Right. We're on the same page. I was saying if whatever resource the uniform set belongs to isn't pooled (the textures in this case based on what you're saying) it would explain why I didn't see any problems. And no I didn't include it in the minimal examples mainly because I didn't want to recreate the shader since I didn't see anything in the other project but I probably should've and turned it on just to cover all the bases. Sorry about that if that wasn't clear.

Zylann:
I pushed the fix branch. There was more nasty changes to make: in the normalmap task, 5 or more uniform sets were created, but *some of them* were leaking, while others were auto-freed. Had to figure out which ones, and ended up free them all before other RIDs to make things simpler (short-circuits the auto-freeing behavior)

PR #741, opened 2025-03-25
Use `Ref.is_valid``Ref.is_null` and rm a unused variable

beicause:
Fixes compilation with `warnings=extra werror=yes strict_checks=yes`

### DIFF
--- a/meshers/blocky/types/voxel_blocky_type.cpp
+++ b/meshers/blocky/types/voxel_blocky_type.cpp
@@ -450,10 +450,10 @@ bool try_get_attribute_index_from_name(
 }
 
 template <typename T>
-unsigned int get_non_null_count(const StdVector<T> &objects) {
+unsigned int get_non_null_count(const StdVector<Ref<T>> &objects) {
 	unsigned int count = 0;
-	for (const T &obj : objects) {
-		if (obj != nullptr) {
+	for (const Ref<T> &obj : objects) {
+		if (obj.is_valid()) {
 			++count;
 		}
 	}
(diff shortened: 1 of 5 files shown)


Zylann:
What happened with `==` and `Ref<T>`? I recall it was possible to compare that way

meshers/blocky/types/voxel_blocky_type.cpp
```diff
@@ -450,10 +450,10 @@ bool try_get_attribute_index_from_name(
 }
 
 template <typename T>
-unsigned int get_non_null_count(const StdVector<T> &objects) {
+unsigned int refs_get_non_null_count(const StdVector<T> &objects) {
```
Zylann, line 453: Why does this need to be renamed?

beicause:
> What happened with `==` and `Ref<T>`? I recall it was possible to compare that way, there are operators for it
> 
> Also... `strict_checks`? I didn't know that before. My very long SCons command line is about to get extra longer I guess... Oh this is ONLY to disallow comparison with null xD why

see https://github.com/godotengine/godot/commit/df29cc696f22e6b91b09284ee9ce0779ed77a3d9

Zylann:
I know this commit, it was just never explained what was the rationale behind such an option. I find it weird because Ref<T> still has comparison operators to compare against `T*` pointers. I dont understand what's up with `nullptr`, maybe "because we want to" I guess

Zylann:
Thanks!

Poikilos:
> I dont understand what's up with `nullptr`, maybe "because we want to" I guess

NULL is int or long, but nullptr is a pointer. So it ensures comparing same (pointer) type.

beicause:
This is just a change of code style. There is no difference between comparing `nullptr` and  `Ref.is_null`, `Ref.is_valid`.

PR #749, opened 2025-04-14
Fix unused field warnings when compiling out gpu and instance features

beicause:


Zylann:
Wut, is there a warning for fields of a struct/class that aren't used?
I actually left them on purpose... because doing this starts to put even more `#ifdefs` for no real gain, and some of these I can't even put easily

beicause:
There are indeed some unused private field warnings. If they are left intentionally, we can use [[maybe unused]]. What do you think?

Zylann:
What kind of configuration are you using for warnings?

Zylann:
> If they are left intentionally, we can use [[maybe unused]]. What do you think?

I don't think it's better, as it's still a special change, but without a reason.
I was hoping the compiler wouldn't be as pedantic here, I don't like it, but I guess `#ifdef` is better.

beicause:
>What kind of configuration are you using for warnings?

I compiled with `warnings=extra`, and these are the only warnings. I think we can enable 
`warnings=extra` in CI.


>I don't think it's better, as it's still a special change, but without a reason.
I was hoping the compiler wouldn't be as pedantic here, I don't like it, but #ifdef is better for context.

I will change back to `#ifdef`. Mono build seems to ignore attributes.

Zylann:
> I compiled with warnings=extra

Oh dear, those warnings are incredibly pedantic, no wonder^^" It really doesn't play well with stripping individual features

.github/workflows/extension_builds.yml
```diff
@@ -94,20 +94,20 @@ jobs:
...
           key: ${{ matrix.target.platform }}_${{ matrix.target.arch }}_${{ matrix.float-precision }}_${{ matrix.target-type }}_${{ inputs.godot-cpp-branch }}_gdextension_cache
-      
+
       - name: Build
         shell: sh
         env:
           SCONS_CACHE: ${{ github.workspace }}/.scons-cache/
         run: |
-          scons target=${{ matrix.target-type }} platform=${{ matrix.target.platform }} arch=${{ matrix.target.arch }} precision=${{ matrix.float-precision }} dev_build=no verbose=yes
+          scons target=${{ matrix.target-type }} platform=${{ matrix.target.platform }} arch=${{ matrix.target.arch }} precision=${{ matrix.float-precision }} dev_build=no verbose=yes warnings=extra
```
Zylann, line 110: You can't use that here. That option doesn't exist in GodotCpp.

.github/workflows/extension_strip_checks.yml
```diff
@@ -65,4 +65,4 @@ jobs:
         env:
           SCONS_CACHE: ${{ github.workspace }}/.scons-cache/
         run: |
-          scons target=editor platform=linux arch=x86_64 dev_build=no verbose=yes use_llvm=yes ${{ matrix.sconsflags }}
+          scons target=editor platform=linux arch=x86_64 dev_build=no verbose=yes warnings=extra use_llvm=yes ${{ matrix.sconsflags }}
```
Zylann, line 68: You can't use that here. That option doesn't exist in GodotCpp.

Zylann (commented):
The GodotCpp strip check

PR #751, opened 2025-04-15
Regenerate classes doc

beicause:


Zylann:
This doesn't update anything. In fact, latest Godot has an annoying habit of removing the "old way" some options worked, but by doing so, it breaks them when the module is compiled in older versions of Godot. So the way it works currently is that if the "old" way still works, it remains used. If not, both old and new ways are used.

beicause:
This change of document seems to have started from godot 4.3 https://github.com/godotengine/godot/commit/af28f87791ac4aed15d8f869876296febf5d5b58. voxel module appears to support 4.0+, so closing.

PR #754, opened 2025-04-22
FastNoise2: Add simd support for Android

beicause:
Tested on arm64 Android phone, Neon works.
Not yet test arm32, but I guess it should also work.

### DIFF
--- a/SCsub
+++ b/SCsub
@@ -1,4 +1,3 @@
-
 # This is the entry point for SCons to build this engine as a module.
 # It has to be integrated to Godot's codebase, so it cannot be run on its own, see `README` for more information.
 # To build as an extension, see `SConstruct`.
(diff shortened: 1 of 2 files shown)


SCsub
```diff
@@ -11,82 +10,82 @@ import voxel_version
...
+    env_voxel.Append(CPPDEFINES="TRACY_ENABLE")
+    env_voxel.Append(CPPPATH=["#thirdparty/tracy/public"])
+    voxel_files += ["#thirdparty/tracy/public/TracyClient.cpp"]
 
 # ----------------------------------------------------------------------------------------------------------------------
-
-for f in voxel_files:
-	env_voxel.add_source_files(env.modules_sources, f)
+module_obj = []
+env_voxel.add_source_files(module_obj, voxel_files)
```
Zylann, line 81: Why this change?

SCsub
```diff
@@ -11,82 +10,82 @@ import voxel_version
...
+env.modules_sources += module_obj
 
 # TODO Feature: check webassembly builds (`env["platform"] == "javascript"`)
 
 # Doesn't work, since the rest of Godot doesn't use this, linking fails.
 # No safe STL boundary checks for you.
-#if env["target"] == "debug":
-#	if env.msvc:
-#		# Enable STL bound checks, Godot's master environment doesn't do it
-#		env_voxel.Append(CXXFLAGS=["/D_DEBUG"])
```
Zylann, line 92: This has lost indentation information

SCsub
beicause: https://github.com/godotengine/godot/blob/0ed1c192e87e107f507d41dae00f9859bcc45ef1/methods.py#L34

This style is also used in godot modules, which avoids multiple string checks.

SCsub
```diff
@@ -11,82 +10,82 @@ import voxel_version
...
+        env_voxel.Append(CXXFLAGS=["-Wno-unknown-pragmas"])
+
+    # Build from source. Should be the simplest, but requires C++17
+    SConscript("thirdparty/fast_noise_2/SConscript", exports=["env", "env_voxel"])
+
+    env_voxel.Append(CPPPATH=["thirdparty/fast_noise_2/include"])
+
+    voxel_files += ["util/noise/fast_noise_2.cpp"]
+
+    if env.editor_build:
```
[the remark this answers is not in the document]
Zylann: But the code from before was already working with it. Now there cant be globs, I dont understand the point.

SCsub
Zylann: This for loop was used so that glob strings could be used. Now they can't...
Also not all modules do this, and those I've seen actually need the `module_obj` to pass them to `env.Depends`, so I thought that's why things were done this way?

SCsub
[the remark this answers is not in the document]
beicause: Re-add the globs now.

beicause:
There is still a issue that SSE is also compiled on arm unnecessarily. Meanwhile, I didn't consider X86 Android. I will fix these later.
https://github.com/Zylann/godot_voxel/blob/56e4cf1d4c9cced521552367ffc32d5d2f9ae1b8/thirdparty/fast_noise_2/SConscript#L133-L137

beicause:
I think we should compile FastNoise2 on all platforms by default. At least the scalar level is always available.

beicause:
I don't have an iOS device, but I created a branch to test iOS in the CI, and it also compiled successfully. https://github.com/beicause/godot_voxel/actions/runs/14616030217/job/41004367330
So NEON might work in arm macOS and iOS ?

Zylann:
Needs to be tested, I don't have a Mac to confirm that.

PR #768, opened 2025-06-09
Allow to set thread count at runtime

wildlachs:
This PR exposes two new methods in the `VoxelEngine` singleton (the GDScript version of the class):

- `int get_thread_count()`:
Returns the number of currently used threads in the thread pool.
Implemented as a call chain:
```
zylann::voxel::godot::VoxelEngine.get_thread_count()
  -> zylann::voxel::VoxelEngine.get_thread_count()
    -> _general_thread_pool.get_thread_count()
```

- `void set_thread_count(int count)`:
Sets the number of threads to be used by the thread pool.
Also implemented as a call chain:
```
zylann::voxel::godot::VoxelEngine.set_thread_count(int count)
  -> zylann::voxel::VoxelEngine.set_thread_count(int count)
    -> _general_thread_pool.set_thread_count(int count)
```

The docs have been updated to include these two new methods.

This PR also includes a fix for how the `ThreadedTaskRunner` handles a change of thread count.

I have tested this both with the official [voxelgame](https://github.com/Zylann/voxelgame) demo and my personal project.
It doesn't seem any tasks get dropped, so the internal logic of `destroy_all_threads()` with subsequent calls to `create_thread(d, i)` appears to be robust. However, yielding all threads can cause noticeable stuttering, as is described in the docs.
Depending on the use case, this might be negligible though.

Motivation:
The new methods added by this PR perfectly complement a terrain pre-generation approach using a roamer: The thread count can be set to a very high value on first generation, then the roamer can be run, and finally the thread count can be lowered again to allow fluid gameplay.

### DIFF
--- a/doc/classes/VoxelEngine.xml
+++ b/doc/classes/VoxelEngine.xml
@@ -42,6 +42,19 @@
 				[/codeblock]
 			</description>
 		</method>
+		<method name="get_thread_count" qualifiers="const">
+			<return type="int" />
+			<description>
+				Returns the number of threads currently used internally by the [code]ThreadedTaskRunner[/code].
+			</description>
+		</method>
+		<method name="set_thread_count">
+			<return type="void" />
+			<param index="0" name="count" type="int" />
+			<description>
+				Sets the number of threads to be used internally by the [code]ThreadedTaskRunner[/code]. Setting this can cause lagging, and it might take some time until the number of threads actually matches the given value.
+			</description>
+		</method>
 		<method name="get_threaded_graphics_resource_building_enabled" qualifiers="const">
 			<return type="bool" />
 			<description>
--- a/engine/voxel_engine.cpp
+++ b/engine/voxel_engine.cpp
@@ -411,7 +411,7 @@ int VoxelEngine::get_thread_count() const {
 	return _general_thread_pool.get_thread_count();
 }
 
-void VoxelEngine::set_thread_count(int count) {
+void VoxelEngine::set_thread_count(uint32_t count) {
 	_general_thread_pool.set_thread_count(count);
 }
 
--- a/engine/voxel_engine.cpp
+++ b/engine/voxel_engine.cpp
@@ -407,4 +407,12 @@ VoxelEngine::Stats VoxelEngine::get_stats() const {
 	return s;
 }
 
+int VoxelEngine::get_thread_count() const {
+	return _general_thread_pool.get_thread_count();
+}
+
+void VoxelEngine::set_thread_count(int count) {
+	_general_thread_pool.set_thread_count(count);
+}
+
 } // namespace zylann::voxel
--- a/engine/voxel_engine_gd.cpp
+++ b/engine/voxel_engine_gd.cpp
@@ -177,6 +177,19 @@ Dictionary VoxelEngine::get_stats() const {
 	return to_dict(zylann::voxel::VoxelEngine::get_singleton().get_stats());
 }
 
+int VoxelEngine::get_thread_count() const {
+	return zylann::voxel::VoxelEngine::get_singleton().get_thread_count();
+}
+
+void VoxelEngine::set_thread_count(int count) {
+	ERR_FAIL_COND_MSG(count < 1,
+			vformat("The thread count must be a number from 1 to %d", ThreadedTaskRunner::MAX_THREADS));
+	// if `thread_count` is bigger than MAX_THREADS we allow the function to continue,
+	// as the rest of the code is build to handle this
+
+	zylann::voxel::VoxelEngine::get_singleton().set_thread_count(count);
+}
+
 void VoxelEngine::schedule_task(Ref<ZN_ThreadedTask> task) {
 	ERR_FAIL_COND(task.is_null());
 	ERR_FAIL_COND_MSG(task->is_scheduled(), "Cannot schedule again a task that is already scheduled");
@@ -235,6 +248,8 @@ void VoxelEngine::_bind_methods() {
 	ClassDB::bind_method(D_METHOD("get_version_status"), &VoxelEngine::get_version_status);
 	ClassDB::bind_method(D_METHOD("get_version_git_hash"), &VoxelEngine::get_version_git_hash);
 	ClassDB::bind_method(D_METHOD("get_stats"), &VoxelEngine::get_stats);
+	ClassDB::bind_method(D_METHOD("get_thread_count"), &VoxelEngine::get_thread_count);
+	ClassDB::bind_method(D_METHOD("set_thread_count", "count"), &VoxelEngine::set_thread_count);
 
 	ClassDB::bind_method(
 			D_METHOD("get_threaded_graphics_resource_building_enabled"),
--- a/engine/voxel_engine_gd.cpp
+++ b/engine/voxel_engine_gd.cpp
@@ -181,10 +181,11 @@ int VoxelEngine::get_thread_count() const {
 	return zylann::voxel::VoxelEngine::get_singleton().get_thread_count();
 }
 
-void VoxelEngine::set_thread_count(uint32_t count) {
-	ERR_FAIL_COND_MSG(count < 1 || count > ThreadedTaskRunner::MAX_THREADS,
-			vformat("Thread count must be a number from 1 to %d", ThreadedTaskRunner::MAX_THREADS));
-	zylann::voxel::VoxelEngine::get_singleton().set_thread_count(count);
+void VoxelEngine::set_thread_count(int count) {
+	constexpr int MAX_THREADS = static_cast<int>(ThreadedTaskRunner::MAX_THREADS);
+	ERR_FAIL_COND_MSG(count < 1 || count > MAX_THREADS,
+			vformat("Thread count must be a number from 1 to %d", MAX_THREADS));
+	zylann::voxel::VoxelEngine::get_singleton().set_thread_count(static_cast<uint32_t>(count));
 }
 
 void VoxelEngine::schedule_task(Ref<ZN_ThreadedTask> task) {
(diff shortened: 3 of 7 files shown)


engine/voxel_engine_gd.cpp
```diff
@@ -177,6 +177,19 @@ Dictionary VoxelEngine::get_stats() const {
...
 
+int VoxelEngine::get_thread_count() const {
+	return zylann::voxel::VoxelEngine::get_singleton().get_thread_count();
+}
+
+void VoxelEngine::set_thread_count(int count) {
+	ERR_FAIL_COND_MSG(count < 1,
+			vformat("The thread count must be a number from 1 to %d", ThreadedTaskRunner::MAX_THREADS));
+	// if `thread_count` is bigger than MAX_THREADS we allow the function to continue,
+	// as the rest of the code is build to handle this
```
Zylann, line 188: Except it is silently handled, and it is preferable to not assume that, you should put an error check.

doc/classes/VoxelEngine.xml
```diff
@@ -42,6 +42,19 @@
...
+			<return type="int" />
+			<description>
+				Returns the number of threads currently used internally by the [code]ThreadedTaskRunner[/code].
+			</description>
+		</method>
+		<method name="set_thread_count">
+			<return type="void" />
+			<param index="0" name="count" type="int" />
+			<description>
+				Sets the number of threads to be used internally by the [code]ThreadedTaskRunner[/code]. Setting this can cause lagging, and it might take some time until the number of threads actually matches the given value.
```
Zylann, line 55: There are ways to have no lag. I just dont clearly remember why `set_thread_count` destroys all the threads. Maybe there isn't a particular reason other than it was used only in the destructor. But maybe for now it can be with lag.

engine/voxel_engine.cpp
```diff
@@ -407,4 +407,12 @@ VoxelEngine::Stats VoxelEngine::get_stats() const {
 	return s;
 }
 
+int VoxelEngine::get_thread_count() const {
+	return _general_thread_pool.get_thread_count();
+}
+
+void VoxelEngine::set_thread_count(int count) {
```
Zylann, line 414: Should take a uint

util/tasks/threaded_task_runner.cpp
```diff
@@ -90,11 +90,11 @@ void ThreadedTaskRunner::set_thread_count(uint32_t count) {
 		count = MAX_THREADS;
 	}
 	destroy_all_threads();
-	for (uint32_t i = _thread_count; i < count; ++i) {
+	_thread_count = count;
+	for (uint32_t i = 0; i < _thread_count; ++i) {
```
Zylann, line 94: I have a feeling that method was partially thought to be implemented incrementally, without destroying all threads. But I must have reverted to destroy them anyways because I didnt bother changing at runtime. Yet part of the incremental way remained.
Adding more threads means creating the extra ones.
Removing threads means "flagging" the extra ones so that they stop by themselves.
Though it seems a bit easier to destroy them all, at the cost of lag.

Zylann:
> I have also changed the arguments to uint32_t. Is this correct, or might GDScript cause an overflow if I try to call set_thread_count with a really big number?

The idea is to keep using `int` on the method you bind, so the error check will be more clear to users than an unsigned underflow.

engine/voxel_engine_gd.cpp
```diff
@@ -177,6 +177,16 @@ Dictionary VoxelEngine::get_stats() const {
 	return to_dict(zylann::voxel::VoxelEngine::get_singleton().get_stats());
 }
 
+int VoxelEngine::get_thread_count() const {
+	return zylann::voxel::VoxelEngine::get_singleton().get_thread_count();
+}
+
+void VoxelEngine::set_thread_count(uint32_t count) {
```
Zylann, line 184: Don't change the exposed method to use `uint32_t`, keep it as `int` and check if negative.

wildlachs:
The exposed function now takes an `int`, while the internal functions keep `uint32_t` argument types.

In the meantime I was able to do some more testing.
Here are a few of my observations that I think are worth mentioning:

- The editor build uses very un-optimized synchronization primitives (probably for a good reason though). If you are using a `VoxelGeneratorScript` the GDScript runtime will heavily fight for `SpinLock`s when invoked from multiple threads. This is so extreme that – at least in my tests – running generation on a single thread is as fast as using 2, 3, or 4 threads. When using more than 4 threads performance starts to drop. When I ran generation with 31 threads (on my 32 thread processor) generation was *significantly* slower compared to a single thread (at least 20x lower).

- The release build uses far more optimized synchronization primitives. Even when running 31 threads, I have barely seen any locking in the flamegraph. *By far* the biggest performance cost comes from `free` operations (which are required since resources are being allocated for running the generation threads).

- If your game has a system that keeps track of block (un)loading, which it definitely should have, you can actually use `set_thread_count()` to dial in the perfect amount of multithreading at runtime. All it takes is a little function measuring the number of accomplished tasks within a given time frame, and now you got a metric to check for when in-/decreasing thread count. It would be nice for this if it wouldn't cause lag though.

- If you are not using `VoxelGeneratorScript` you can reach absolutely incredible levels of performance. Using my 5900XT I was able to fly through the blocky_terrain demo at 1000 blocks/s and it just kept generating...

Zylann:
Hmm that's quite a lengthy post to add to such a simple PR don't you think xD (especially since doing all these benchmarks could have been done independently from this PR, or without it)
I can't see the videos, none of them are loading for me.

>  the thread pool can generate blocks faster than the VoxelTerrain can consume it.

Quick note, physics is a bottleneck. If your project requires Godot Physics colliders, they can't be baked on threads, so they are done on the main thread which has an abysmal impact on throughput. This has been a problem for years going all the way back to Godot 3 times and remains a problem with the Jolt integration.

>  I know that the biggest impact on performance comes from the fact that you have to (re-)allocate memory. GDScript is almost like a textbook example for this

How so? Where is GDScript allocating memory? I was more under the impression GDScript was essentially fighting for debug locks due to the sheer amount of script functions running in parallel.
What is true however is that there is a lot of voxel data to fill in and that can indeed put pressure on memory access. But not sure if that's something you notice first with a GDScript benchmark?

> My interpretation is that the memory controller has a hard time feeding the CPU cores – caused by a lot of memory (re-)allocations – caused by GDScript being essentially operations on a stack.

I can't see the video and not really understanding this, did you verify that it's all about GDScript being slow? Is this in release?

I also want to point out you made this benchmark on a specific machine, but it will be different on many others. The point of the project setting originally was to have an option to automatically scale without having to do super-deep research on every CPU on the market to run game-specific benchmarks, which may not be as important as many other areas of development. But I guess if you have the time to do that, you can force a different count now.

wildlachs:
> Hmm that's quite a lengthy post to add to such a simple PR don't you think xD (especially since doing all these benchmarks could have been done independently from this PR, or without it)

> Might have to bookmark this in case I need this reference later.

Absolutely! I have planned to some more performance testing (and debugging) once 4.5 comes out (if ... it ... eventually ... comes out). I'll post my results in a separate issue then.

> Quick note, physics is a bottleneck. If your project requires Godot Physics colliders, they can't be baked on threads, so they are done on the main thread which has an abysmal impact on throughput. This has been a problem for years going all the way back to Godot 3 times and remains a problem inherent to the Jolt integration.

Yup, my game uses physics collision, and I am painfully aware of that.
But let's just say, there might be some more PRs coming down the road for this. (After all, for a terrain generation engine, it makes sense to have multiple "logical" LOD ranges, right? And not all of them might need all features – like collision generation. But that's a topic for the future...)

> Also yes there will eventually be a "sweet spot" depending on how your project is setup, but other factors might influence it, such as what CPU your players will have (which can vary) and whether other threads in Godot might compete in the background (Godot also has its own thread pool for some things, which works slightly differently).

Exactly, that's why testing is so important. And personally, I am a big fan of systems that can adapt to their environment on-the-fly.

> How so? Where is GDScript allocating memory? I was more under the impression GDScript was essentially fighting for debug locks due to the sheer amount of script functions running in parallel. It is also indeed an interpreter, but just running code doesnt allocate memory dynamically, isnt it? (I hope not)

Kinda depends. When it comes to fighting for locks: In debug builds, absolutely! It's horrible. But in release builds, almost all of that stuff gets opted-out (thank god). However, this makes it almost impossible to do debugging/testing (trust me, DWARF won't save you here).
But yeah, it still is an **interpreted** language. And it's great for its intended purpose! That is: fast prototyping, light to medium heavy scripting, duck typing (quack), and seamless integration with Godot internals.
But it sucks at heavy performance lifting, because it is essentially performing operations on a stack. That means: no registers (forget CFFI, forget fast access to function args, forget return value optimization, etc.), constantly calling into virtual function tables (because you are `switch`ing on instruction values), a bunch of unneeded PUSH/POP operations (which, in a multithreaded context, might even require synchronization by the CPU! No fast L1/L2 cache for you dangit...), nothing that comes even remotely close to respecting the fact that *cache lines exist*, and so on.

If you want to get a grasp of this, I recommend taking a look at `modules/gdscript/gdscript_vm.cpp`, especially at the implementation of `GDScriptFunction::call`. There is a loop in there that is **HOT**. And it contains tons of memory-heavy operations.
If you want to learn more about how interpreters work under the hood, I can highly recommend CRAFTING INTERPRETERS by Robert Nystrom. Really one of the better books I have worked my way through in the last 5 years. (It can also be read online completely for free)

> did you verify that it's all about GDScript being slow? Is this in release?

I mean, in blocky_game and a release build `_generate_block` was actually quite fast. Easily reaching 30.000 blocks per second (that is over 120 billion voxels per second) as can be seen in the plots. In my personal game, the generator has **a lot** more entanglements with different parts of the game (almost all GDScript so far). I am currently porting parts of the generator over to C++, we'll see how much that improves performance. I'll include my findings in the proper issue (when 4.5 finally comes out).

Thanks for merging!

PR #794, opened 2025-09-17
Fix uniform blocks getting skipped by `run_blocky_random_tick`

NuclearPhoenixx:
Just a quick and easy fix to uniform blocks getting skipped by calls to `run_blocky_random_tick` despite them having a `VoxelBlockyModel` which is tickable. This probably must have reversed the entire process instead, so uniform blocks that were **not** set to tickable were actually called? Not sure, but now uniform blocks are random ticked as expected. Thanks to @Zylann!


Zylann:
Thanks!

PR #796, opened 2025-09-17
Update meshoptimizer to v0.25

NuclearPhoenixx:
Updated from somewhere around [v0.16](https://github.com/zeux/meshoptimizer/releases/tag/v0.16) (4 years ago) to the latest version as of now: [v0.25](https://github.com/zeux/meshoptimizer/releases/tag/v0.25)

Compiled with the latest master branch and Godot 4.5 stable, no issues there. A quick test with a bare VoxelTerrain, the trans voxel mesher and a noise generator everything looked alright to me. Maybe a few other people can test this with their projects just to be sure nothing broke.

One notable change with the new version was the addition of an `options` bitmask to the `meshopt_simplify` function. In the documentation it mentions `options must be a bitmask composed of meshopt_SimplifyX options; 0 is a safe default` so I left it at 0 for now.

<details>

https://github.com/Zylann/godot_voxel/blob/b1b1a4e79090c5586b63aca417d5e8071bd07b06/thirdparty/meshoptimizer/meshoptimizer.h#L461

https://github.com/Zylann/godot_voxel/blob/b1b1a4e79090c5586b63aca417d5e8071bd07b06/meshers/transvoxel/voxel_mesher_transvoxel.cpp#L180-L192
</details>

PR #798, opened 2025-09-18
Meshoptimizer: Remove custom implementation

NuclearPhoenixx:
**This PR supersedes #796.**

Since the issue with seams in terrain chunk meshes (https://github.com/zeux/meshoptimizer/issues/311) has been fixed by the meshoptimizer library with [v0.18](https://github.com/zeux/meshoptimizer/releases/tag/v0.18), it is no longer necessary to include these changes in a custom copy of the lib for this module.

This PR removed the entire custom meshoptimizer thirdparty lib from the module, and switches to use the built-in lib from Godot for the transvoxel mesher by using the `meshopt_SimplifyLockBorder` option to fix chunk borders. This reduces duplicate code and removes the additional namespace making maintenance easier for the future.

This should not break compatability with any Godot 4.x versions since these changes to meshoptimizer were backported all the way to Godot 4.0 in the past.

Tested with the latest master branch and Godot 4.5, no issues have been found.

---

**Terrain with default Godot meshoptimizer and options bitmask `0`:**

screenshot_20250918_122250

**Terrain with default Godot meshoptimizer and options bitmask `meshopt_SimplifyLockBorder`:**

screenshot_20250918_122413

### DIFF
--- a/edition/funcs.cpp
+++ b/edition/funcs.cpp
@@ -208,7 +208,7 @@ void run_blocky_random_tick(
 					const uint64_t v = voxels.get_voxel(0, 0, 0, channel);
 					if (lib_data.has_model(v)) {
 						const blocky::BakedModel &vt = lib_data.models[v];
-						if (vt.is_random_tickable) {
+						if (!vt.is_random_tickable) {
 							// Skip whole block
 							continue;
 						}
--- a/thirdparty/lz4/VERSION.txt
+++ b/thirdparty/lz4/VERSION.txt
@@ -1,0 +1,4 @@
+lz4/lz4
+
+v1.10.0 (ebb370ca83af193212df4dcbadcc5d87bc0de2f0)
+Jul 22, 2024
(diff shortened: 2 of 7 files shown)


NuclearPhoenixx:
Also, issue https://github.com/zeux/meshoptimizer/issues/158 was apparently also fixed.

https://github.com/Zylann/godot_voxel/blob/03a8615a6047627b917ca1a5f422349bfcaaedbf/meshers/transvoxel/voxel_mesher_transvoxel.cpp#L335-L337

And https://github.com/zeux/meshoptimizer/issues/312 seems to have also been fixed?

https://github.com/Zylann/godot_voxel/blob/03a8615a6047627b917ca1a5f422349bfcaaedbf/meshers/transvoxel/transvoxel.cpp#L1122

NuclearPhoenixx:
Hm, looks like I broke most of the actions builds 🤔

Zylann:
You broke extension builds because meshoptimizer is not available there. So the library is still needed.
I would also not assume it is available in module builds when compiling export templates, I remember it's used on mesh import which is an editor-only process?

PR #800, opened 2025-09-19
Fix compilation with `voxel_smooth_meshing=no`

beicause:


NuclearPhoenixx:
What does this do?

Zylann:
`voxel_smooth_meshing=no` is an option that compiles the module without smooth meshing. The idea was to reduce a bit the size of the executable (not by a lot, but it's something)

PR #856, opened 2026-04-13
Improve `VoxelBoxMover` stepping when sliding

### LINKED ISSUE #813: `VoxelBoxMover`: Unable to step when sliding along wall
**Describe the bug**
When sliding on a wall, you cannot step up another block despite the step height being smaller than the max step height.

**To Reproduce**
1. Set voxel box mover max step height to 0.5
2. Place a few blocks with height 1 in front of the player
3. Place a block with height 0.5 to the side
4. Now slide along the wall, and you won't step up the half block

I've tried around to fix it, got close but haven't found any solution that makes stepping reliable with the sliding as well, but I'm also pretty much at my limit here.

https://github.com/NuclearPhoenixx/godot_voxel/commit/375dedc70d1c49241275f43ee5bd8c6180f74d2d

**Expected behavior**
Sliding along blocks should not interfere with the stepping mechanic.

**Screenshots**

https://github.com/user-attachments/assets/539352d0-0b5f-491b-ad2b-2ee84172d041

**Environment**
Latest godot version from the stable branch and godot_voxel from the master branch.

NuclearPhoenixx:
Should pretty much fix https://github.com/Zylann/godot_voxel/issues/813. I checked it in my demo world from the issue and it works pretty much perfectly now. Stepping free-standing slabs works as expected, and also the collision checks during stepping seem to still work perfectly when there are other voxels (maybe partially) blocking the step.

It comes at a cost, though, I'm pretty much doing it all the "lazy" way by using `zylann::voxel::get_motion` 3x and letting it do most of the work, which I'm sure is worse for performance, but I don't think it'll be that noticeable in any case. And I think this way we shouldn't run in any weird edge cases where stepping glitches, but it doesn't hurt if someone elses tests it too!

### DIFF
--- a/terrain/fixed_lod/voxel_box_mover.cpp
+++ b/terrain/fixed_lod/voxel_box_mover.cpp
@@ -323,6 +323,15 @@ Vector3 VoxelBoxMover::get_motion(
 
 		// Only step if we're blocked horizontally, i.e. the actual motion is less than the intended
 		if (actual_h_motion.length_squared() < intended_h_motion.length_squared() - epsilon) {
+			// Actually forgot to get new collision boxes from the terrain for the "step moved" position.
+			// If we don't do this, we will glitch through all solid voxels above us when stepping.
+			AABB step_temp_box = box;
+			step_temp_box.size.y += _max_step_height;
+			step_temp_box = expand_with_vector(step_temp_box, motion);
+
+			potential_boxes.clear();
+			collect_boxes(terrain_data, mesher, step_temp_box, _collision_mask, potential_boxes);
+
 			// Basically we're now going to manually push the player up to the max step height, check for the possible horizontal motion,
 			// push the player back down to the "real" step height of the terrain, and then check if we've gotten farther than before.
 			AABB step_box = box; // Starting box
--- a/terrain/fixed_lod/voxel_box_mover.cpp
+++ b/terrain/fixed_lod/voxel_box_mover.cpp
@@ -316,42 +316,42 @@ Vector3 VoxelBoxMover::get_motion(
 	// Minecraft-style stair climbing:
 	// If we were moving, changed horizontal direction due to collision, and resulting motion is about horizontal
 	_has_stepped_up = false;
-	if (_step_climbing_enabled &&
-		// Movement is horizontal?
-		Math::abs(slided_motion.y) < 0.001 && Vector2(motion.x, motion.z).length_squared() > 0.0001 &&
-		// Motor movement isn't the same as resulting slided motion?
-		Vector2(motion.x, motion.z).normalized().dot(Vector2(slided_motion.x, slided_motion.z).normalized()) < 0.99) {
-		// We hit an obstacle
-		real_t hit_y;
-		// Find out the height of the step
-		if (boxcast_down(to_span(potential_boxes), get_xz(expanded_box.position), get_xz(expanded_box.size), hit_y)) {
-			// If the step is up and not too high
-			if (hit_y > box.position.y && (hit_y - box.position.y) <= _max_step_height) {
-				// Check if we would fit if we move the box above the step.
-				// Raise it slightly higher to avoid precision issues. Even if the final motion would move the box
-				// exactly on top of the stair, gameplay code could do some additional calculations with that motion
-				// (converting it to velocity?) which may induce precision errors causing the box to fall through.
-				const real_t epsilon = 0.0001f;
-				const AABB hyp_box(
-						Vector3(box.position.x + motion.x, hit_y + epsilon, box.position.z + motion.z), box.size
-				);
+	const real_t epsilon = 0.0001f;
+	if (_step_climbing_enabled && Math::abs(motion.y) <= 0.001 && Vector2(motion.x, motion.z).length_squared() > epsilon) {
+		Vector2 intended_h_motion(motion.x, motion.z);
+		Vector2 actual_h_motion(slided_motion.x, slided_motion.z);
 
-				potential_boxes.clear();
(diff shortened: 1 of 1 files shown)


Zylann:
Did you check the https://github.com/Zylann/godot_voxel/tree/step_climb2 branch?

NuclearPhoenixx:
Yes, but somehow it actually didn't change anything, the result looks the same as in my testing video unfortunately.

Zylann:
Argh, that's really odd because it did change something for me, spent a bunch of time figuring it out.
Testing it again it appears it indeed stops working but only if I lower the speed the character goes at in the demo, which is... also weird, because speed isn't supposed to have such an impact on stair climbing 🤔

Zylann:
That was an [epsilon](https://github.com/Zylann/godot_voxel/commit/908ce316f981bbedf9c0fc4f444cf87c5aad0bcd) problem: 
[step_climb2.webm](https://github.com/user-attachments/assets/a21c8e3e-7993-4976-9015-2c53a32d43fd)

NuclearPhoenixx:
That works fine now, but there is still one issue left. Somehow the collision detection gets confused with a voxel above the player and it looks like it wants to step, but can't. In this demo there is a voxel above the player, and you can see the small jumps, that's not me, that's just the stepping doing its thing. I don't see this effect in my code.

https://github.com/user-attachments/assets/50a96c99-4043-46a6-be56-dbfc02cf6501

Zylann:
Sorry I can't see the full configuration you have in this video, are you on a slab? Where is the voxel above?

NuclearPhoenixx:
Yeah I'm on a slab, and there is basically 1,5 voxels of space until there are two slabs above (can probably be any voxel). It's the same setup as before just with the two voxels above. And it seems somehow (I guess) that it wants to step up on the full stone voxel, but the voxel above is blocking it, which is what should happen, but I'm not sure why it seems to be trying anyways?

screenshot_20260414_190119

I hope it's not something weird related to my player movement code, but then again, I think it should be the same effectt with the code from this PR.

Zylann:
How tall is your character?

NuclearPhoenixx:
Default for the hitbox y is 1.8, when crouching its 1.44 (80%), so in the video above it was 1.44.

Zylann:
Found another edge case I fixed in https://github.com/Zylann/godot_voxel/commit/a8304be4e4107a4d1622b26cc239e7c36d04db3a
Then I tested your case and could not reproduce:
[step_climb2_case2.webm](https://github.com/user-attachments/assets/c05f9f28-b3dd-4cdd-b551-f960c7be5755)
(why are my videos not embedding 🤔 )

NuclearPhoenixx:
Huh, ok weird, I'll look into it more.

> (why are my videos not embedding 🤔 )

webm maybe?

Zylann:
> webm maybe?

How ironic for a web format

NuclearPhoenixx:
> Found another edge case I fixed in [a8304be](https://github.com/Zylann/godot_voxel/commit/a8304be4e4107a4d1622b26cc239e7c36d04db3a) Then I tested your case and could not reproduce: [step_climb2_case2.webm](https://github.com/user-attachments/assets/c05f9f28-b3dd-4cdd-b551-f960c7be5755) (why are my videos not embedding 🤔 )

Does it also fire `VoxelBoxMover.has_stepped_up()` for you when sliding along the edge with the ceiling above?

I just checked that when I do this, I get a slight +y motion of 0.0001 every (physics) frame when walking against that edge with the ceiling and then after a few seconds I get a single ~ -0.007 movement down in y and at the same time this jumping effect, which is probably the physics interpolation not liking the slow and steady increase and then the reset down all at once.

And yes, during this whole thing, `has_stepped_up` is firing, although I'm obviously not stepping anywhere.

Funnily, enough I just checked and I get the exact same even when walking against just a simple "full" voxel. Like a 1x1x1 cube and the player `max_step_height` of 0.51, I get this little +y motion and then the small "jump" reset effect, with `has_stepped_up` being true all the time just by walking against the block (despite not actually stepping ofc). Very weird...

Zylann:
> Does it also fire VoxelBoxMover.has_stepped_up() for you when sliding along the edge with the ceiling above?

Yes it sets the boolean behind `has_stepped_up()`: https://github.com/Zylann/godot_voxel/blob/a8304be4e4107a4d1622b26cc239e7c36d04db3a/terrain/fixed_lod/voxel_box_mover.cpp#L393

I hate floats and physics so much.  I suppose this is yet another epsilon of some sort because when you get blocked by anything in your path, of course the stepping code will try to run, so that includes a simple wall too. Then it passes `if (slided_motion2.length_squared() > slided_motion1.length_squared()) {` even if the difference is a tiny unnoticeable epsilon induced by the calculations. Cuz I didn't notice it tbh. If that tiny effect is bothering you I could add the epsilon to that check.

NuclearPhoenixx:
So yes, simply adding the epsilon from the block above seems to fix it, now everything looks to be working perfectly :)

``if (slided_motion2.length_squared() > slided_motion1.length_squared() + epsilon)``

NuclearPhoenixx:
By the way, in the issue you mentioned

> I thought I was onto something and then it comes up worse and full of edge cases.

I haven't noticed anything, at least for my usecase it looks perfectly fine. Is there anything else that I should test?

Zylann:
That might have been my state of mind before I make the branch with the newer stuff.

NuclearPhoenixx:
Should be good to merge into master then, right?

Zylann:
I just merged `step_climb2` to master.
Sorry for the delay, I was abroad for the GodotCon this week.

NuclearPhoenixx:
No worries, thanks a lot!

PR #861, opened 2026-05-26
Update README.md

kudaktalkaSuperstrong:


Zireael07:
This is mostly workflow changes and removes one empty line from README?

Poikilos:
You deleted the line `branches: [ master, github_actions ]` and `branches: [ master ]`. You basically are not using the workflow as intended (to manage pull requests and ensure master branch has quality control). If you want to run workflows on other branches, I'm not sure why because you can just instead type whatever command you can type while you are debugging. That's up to you, but when workflows run on the project is up to the maintainers.

PR #866, opened 2026-06-04
Compute shader cache

chalecampb:
Providing a PR at this stage for feedback and discussion, with the understanding that there are a few things to test (see test section). It's a pretty neat performance improvement so I hope it's worth merging. 

**Problem Statement** 
When developing the script graph generator I couldn't get it to start rendering before around 5-8s after start. All of my metrics assumed the shader compile time was fixed and long. 

**Root Cause** 
Godot caches shaders except for compute shaders. The fix should probably go in the Godot engine but it would require new infrastructure. 

**Fix Method** 
The previous method was to use shader_create_from_spirv, which calls an internal method _shader_compile_binary_from_spirv followed by shader_create_from_bytecode. 

The fix does a similar duplication in the utils for _shader_compile_binary_from_spirv  to expose the raw binary using RDShaderSPIRV rather than ShaderStageSPIRVData. Then saves it to a buffer, and checks the cache based on the shader source hash and get_pipeline_cache_uuid() (from the documentation this gates against gpu and driver version changes, invalidating the cache). 

In addition, a new shader cache flag was added to the voxel project settings. 

If the flag is false, load_compute_shader_from_glsl is used as it was. If it is enabled, load_compute_shader_from_glsl_cached is called which does the caching and loading logic. The verbose logs will note when cached materials are loaded. 

In the GDExtension, shader_compile_binary_from_spirv is used directly. 

**Testing Performed**
- Time to "Creating VoxelRD pooled storage buffer 36b" text in the logs dropped from around 8s uncached to 0.75s cached on start. 

- I confirmed it builds in modules and gdextension. Using Godot 4.6. 

- I still need to run the tests, and drop the gdextension in a project to confirm, and check in godot 4.7

**Use of AI** 
I used AI, limited to tab complete and providing search utilities (eg, to explain and compare existing cache methods to align with engine methods for shader caching). Most of the code itself was just shuffled around between the engine and voxel module.

**Use of Other Copyright Materials**
I didn't use any materials that were not in the Godot engine.

### DIFF
--- a/engine/gpu/compute_shader.cpp
+++ b/engine/gpu/compute_shader.cpp
@@ -62,24 +62,27 @@ PackedByteArray load_compute_shader_binary_from_cache(const String &cache_file_p
 
 	Error open_error;
 	Ref<FileAccess> file = zylann::godot::open_file(cache_file_path, FileAccess::READ, open_error);
-	if (file.is_null()) {
-		ZN_PRINT_VERBOSE(format("Could not open VoxelRD compute shader binary cache file {} (error {})", cache_file_path, open_error));
-		return PackedByteArray();
-	}
+	ZN_ASSERT_RETURN_V_MSG(
+			file.is_valid(),
+			PackedByteArray(),
+			format("Could not open VoxelRD compute shader binary cache file {} (error {})", cache_file_path, open_error)
+	);
 
 	const uint64_t length = file->get_length();
-	if (length == 0) {
-		ZN_PRINT_VERBOSE(format("VoxelRD compute shader binary cache file is empty for {}", cache_file_path));
-		return PackedByteArray();
-	}
+	ZN_ASSERT_RETURN_V_MSG(
+			length != 0,
+			PackedByteArray(),
+			format("VoxelRD compute shader binary cache file is empty for {}", cache_file_path)
+	);
 
 	PackedByteArray bytes;
 	bytes.resize(length);
 	const uint64_t read_count = zylann::godot::get_buffer(**file, Span<uint8_t>(bytes.ptrw(), bytes.size()));
-	if (read_count != length) {
-		ZN_PRINT_VERBOSE(format("Could not read full VoxelRD compute shader binary cache file for {}", cache_file_path));
-		return PackedByteArray();
-	}
+	ZN_ASSERT_RETURN_V_MSG(
+			read_count == length,
+			PackedByteArray(),
+			format("Could not read full VoxelRD compute shader binary cache file for {}", cache_file_path)
+	);
 
 	return bytes;
 }
--- a/engine/gpu/compute_shader.cpp
+++ b/engine/gpu/compute_shader.cpp
@@ -43,23 +43,19 @@ String get_voxel_compute_shader_cache_base_dir() {
 	base_dir = Engine::get_singleton()->get_shader_cache_path();
 #endif
 	if (base_dir.is_empty()) {
-		base_dir = "user://shader_cache";
+		base_dir = "user://";
 	}
-	return base_dir;
-}
-
-String get_compute_shader_cache_dir(const String &name) {
-	return get_voxel_compute_shader_cache_base_dir()
-			.path_join(name.validate_filename());
+	return base_dir.path_join("shader_cache");
 }
 
 String get_compute_shader_binary_cache_path(const String &source_hash, const String &device_cache_uuid, const String &name) {
 	
-	return get_compute_shader_cache_dir(name)
+	return get_voxel_compute_shader_cache_base_dir()
+			.path_join(name.validate_filename())
 			.path_join(vformat("%s.%s.bin.cache", source_hash, device_cache_uuid));
 }
 
-PackedByteArray load_compute_shader_binary_from_cache(const String &cache_file_path, const String &shader_name) {
+PackedByteArray load_compute_shader_binary_from_cache(const String &cache_file_path) {
 	if (!zylann::godot::file_exists(cache_file_path)) {
 		return PackedByteArray();
 	}
@@ -67,13 +63,13 @@ PackedByteArray load_compute_shader_binary_from_cache(const String &cache_file_p
 	Error open_error;
 	Ref<FileAccess> file = zylann::godot::open_file(cache_file_path, FileAccess::READ, open_error);
 	if (file.is_null()) {
-		ZN_PRINT_VERBOSE(format("Could not open VoxelRD compute shader binary cache file {} for {}", cache_file_path, shader_name));
+		ZN_PRINT_VERBOSE(format("Could not open VoxelRD compute shader binary cache file {} (error {})", cache_file_path, open_error));
 		return PackedByteArray();
 	}
 
 	const uint64_t length = file->get_length();
 	if (length == 0) {
-		ZN_PRINT_VERBOSE(format("VoxelRD compute shader binary cache file is empty for {}", shader_name));
+		ZN_PRINT_VERBOSE(format("VoxelRD compute shader binary cache file is empty for {}", cache_file_path));
 		return PackedByteArray();
 	}
 
@@ -81,7 +77,7 @@ PackedByteArray load_compute_shader_binary_from_cache(const String &cache_file_p
 	bytes.resize(length);
 	const uint64_t read_count = zylann::godot::get_buffer(**file, Span<uint8_t>(bytes.ptrw(), bytes.size()));
 	if (read_count != length) {
-		ZN_PRINT_VERBOSE(format("Could not read full VoxelRD compute shader binary cache file for {}", shader_name));
+		ZN_PRINT_VERBOSE(format("Could not read full VoxelRD compute shader binary cache file for {}", cache_file_path));
 		return PackedByteArray();
 	}
 
@@ -106,7 +102,7 @@ void save_compute_shader_binary_to_cache(
 	Error open_error;
 	Ref<FileAccess> file = zylann::godot::open_file(cache_file_path, FileAccess::WRITE, open_error);
 	if (file.is_null()) {
-		ZN_PRINT_WARNING(format("Could not open VoxelRD compute shader binary cache file {} for writing", cache_file_path));
+		ZN_PRINT_WARNING(format("Could not open VoxelRD compute shader binary cache file {} for writing (error {})", cache_file_path, open_error));
 		return;
 	}
 
@@ -116,8 +112,6 @@ void save_compute_shader_binary_to_cache(
 } // namespace
 
 Ref<RDShaderSPIRV> compile_compute_shader_spirv_from_glsl(RenderingDevice &rd, String source_text, String name) {
-	// Refactored since both cache and non cache paths need this
-	
 	Ref<RDShaderSource> shader_source;
 	shader_source.instantiate();
 	shader_source->set_language(RenderingDevice::SHADER_LANGUAGE_GLSL);
@@ -142,7 +136,7 @@ Ref<RDShaderSPIRV> compile_compute_shader_spirv_from_glsl(RenderingDevice &rd, S
 	return shader_spirv;
 }
 
-RID load_compute_shader_from_glsl(RenderingDevice &rd, String source_text, String name) {
+RID load_compute_shader_from_glsl(RenderingDevice &rd, String source_text, String name, bool shader_cache_enabled) {
 	ZN_PRINT_VERBOSE(format("Creating VoxelRD compute shader {}", name));
 	// For debugging
 	// {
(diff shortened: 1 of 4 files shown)


engine/gpu/compute_shader.cpp
```diff
@@ -29,6 +35,113 @@ String format_source_code_with_line_numbers(String src) {
 	return dst;
 }
 
+namespace {
+
+String get_voxel_compute_shader_cache_base_dir() {
+	String base_dir;
+#if defined(ZN_GODOT)
+	base_dir = Engine::get_singleton()->get_shader_cache_path();
```
Zylann, line 43: Is this not exposed? How are devs normally supposed to know where to put their cache? What will this function return?
Ideally the behavior between module and extension should not be different.

engine/gpu/compute_shader.cpp
```diff
@@ -29,6 +35,113 @@ String format_source_code_with_line_numbers(String src) {
...
+#endif
+	if (base_dir.is_empty()) {
+		base_dir = "user://shader_cache";
```
Zylann: Note that shader compilation happens in editor too. Where would it go in this case?

engine/gpu/compute_shader.cpp
```diff
@@ -29,6 +35,113 @@ String format_source_code_with_line_numbers(String src) {
 	return dst;
 }
 
+namespace {
+
+String get_voxel_compute_shader_cache_base_dir() {
```
Zylann: "voxel" in the name doesnt seem necessary?

engine/gpu/compute_shader.cpp
```diff
@@ -29,6 +35,113 @@ String format_source_code_with_line_numbers(String src) {
...
+#if defined(ZN_GODOT)
+	base_dir = Engine::get_singleton()->get_shader_cache_path();
+#endif
+	if (base_dir.is_empty()) {
+		base_dir = "user://shader_cache";
+	}
+	return base_dir;
+}
+
+String get_compute_shader_cache_dir(const String &name) {
```
Zylann: I was wondering what this function was used for, then reading the rest, I feel it doesnt need to be on its own, it's only used in `get_compute_shader_binary_cache_path`

engine/gpu/compute_shader.cpp
```diff
@@ -29,6 +35,113 @@ String format_source_code_with_line_numbers(String src) {
...
+
+	Error open_error;
+	Ref<FileAccess> file = zylann::godot::open_file(cache_file_path, FileAccess::READ, open_error);
+	if (file.is_null()) {
+		ZN_PRINT_VERBOSE(format("Could not open VoxelRD compute shader binary cache file {} for {}", cache_file_path, shader_name));
+		return PackedByteArray();
+	}
+
+	const uint64_t length = file->get_length();
+	if (length == 0) {
```
Zylann: Why would it be empty?

engine/gpu/compute_shader.cpp
```diff
@@ -29,6 +35,113 @@ String format_source_code_with_line_numbers(String src) {
...
+		ZN_PRINT_VERBOSE(format("VoxelRD compute shader binary cache file is empty for {}", shader_name));
+		return PackedByteArray();
+	}
+
+	PackedByteArray bytes;
+	bytes.resize(length);
+	const uint64_t read_count = zylann::godot::get_buffer(**file, Span<uint8_t>(bytes.ptrw(), bytes.size()));
```
Zylann, line 80: You could use `to_span`

engine/gpu/compute_shader.cpp
```diff
@@ -29,6 +35,113 @@ String format_source_code_with_line_numbers(String src) {
...
+	if (read_count != length) {
+		ZN_PRINT_VERBOSE(format("Could not read full VoxelRD compute shader binary cache file for {}", shader_name));
```
Zylann: Should it be an error?

engine/gpu/compute_shader.cpp
```diff
@@ -29,6 +35,113 @@ String format_source_code_with_line_numbers(String src) {
...
+			.path_join(name.validate_filename());
+}
+
+String get_compute_shader_binary_cache_path(const String &source_hash, const String &device_cache_uuid, const String &name) {
+	
+	return get_compute_shader_cache_dir(name)
+			.path_join(vformat("%s.%s.bin.cache", source_hash, device_cache_uuid));
+}
+
+PackedByteArray load_compute_shader_binary_from_cache(const String &cache_file_path, const String &shader_name) {
```
Zylann: It seems the name isn't used for anything but verbose logs. I normally wouldn't keep it at this level, but it's not a big deal

engine/gpu/compute_shader.cpp
```diff
@@ -29,6 +35,113 @@ String format_source_code_with_line_numbers(String src) {
...
+		return;
+	}
+
+	const Error dir_error = DirAccess::make_dir_recursive_absolute(cache_file_path.get_base_dir());
+	if (dir_error != OK) {
+		ZN_PRINT_WARNING(format("Could not create voxel compute shader binary cache directory for {}", shader_name));
+		return;
+	}
+
+	Error open_error;
```
Zylann, line 105: `open_error` isn't read, sounds like it should be printed in errors, like dir error

engine/gpu/compute_shader.cpp
```diff
@@ -29,6 +35,113 @@ String format_source_code_with_line_numbers(String src) {
...
+void save_compute_shader_binary_to_cache(
+		const String &cache_file_path,
+		const PackedByteArray &shader_binary,
+		const String &shader_name
+) {
+	if (shader_binary.is_empty()) {
+		return;
+	}
+
+	const Error dir_error = DirAccess::make_dir_recursive_absolute(cache_file_path.get_base_dir());
```
Zylann, line 99: Intrigued that the Godot docs dont explicitely say what happens if folders already exist

engine/gpu/compute_shader.cpp
```diff
@@ -29,6 +35,113 @@ String format_source_code_with_line_numbers(String src) {
...
+		return;
+	}
+
+	zylann::godot::store_buffer(**file, to_span(shader_binary));
+}
+
+} // namespace
+
+Ref<RDShaderSPIRV> compile_compute_shader_spirv_from_glsl(RenderingDevice &rd, String source_text, String name) {
+	// Refactored since both cache and non cache paths need this
```
Zylann: This comment should be removed since it only makes sense in the context of the PR itself

engine/gpu/compute_shader.cpp
```diff
@@ -51,24 +159,48 @@ RID load_compute_shader_from_glsl(RenderingDevice &rd, String source_text, Strin
...
+	save_compute_shader_binary_to_cache(binary_cache_file_path, shader_binary, name);
+
+	RID shader_rid = rd.shader_create_from_bytecode(shader_binary);
+	if (shader_rid.is_valid()) {
+		return shader_rid;
 	}
 
-	// TODO What name should I give this shader? Seems it is used for caching
-	const RID shader_rid = zylann::godot::shader_create_from_spirv(rd, **shader_spirv, name);
+	ZN_PRINT_VERBOSE(format("VoxelRD compute shader binary rejected by RenderingDevice for {}, falling back to SPIR-V", name));
```
Zylann: How can we "fallback to SPIRV", when we just did try to compile with SPIRV?
I feel like there is some redundancy going on, perhaps it should be cleaned up.

engine/gpu/compute_shader.cpp
```diff
@@ -92,7 +224,14 @@ void ComputeShaderInternal::clear(RenderingDevice &rd) {
 void ComputeShaderInternal::load_from_glsl(RenderingDevice &rd, String source_text, String name) {
 	ZN_PROFILE_SCOPE();
 	clear(rd);
-	rid = load_compute_shader_from_glsl(rd, source_text, name);
+
+	ZN_ASSERT(ProjectSettings::get_singleton() != nullptr);
```
Zylann: Loading GLSL is failible so perhaps this should be `ZN_ASSERT_RETURN` at the beginning of the function and not crash everything

engine/gpu/compute_shader.cpp
```diff
@@ -92,7 +224,14 @@ void ComputeShaderInternal::clear(RenderingDevice &rd) {
...
+	const bool shader_cache_enabled = ProjectSettings::get_singleton()->get("voxel/shaders/shader_cache/enabled");
+	if (shader_cache_enabled) {
```
Zylann: I wonder if "caching" shouldn't just be a boolean parameter to reduce repetition

engine/voxel_engine_gd.cpp
```diff
@@ -61,6 +61,8 @@ VoxelEngine::Config VoxelEngine::get_config_from_godot() {
 
 	add_custom_project_setting(Variant::BOOL, "voxel/ownership_checks", PROPERTY_HINT_NONE, "", true, true);
 
+	add_custom_project_setting(Variant::BOOL, "voxel/shaders/shader_cache/enabled", PROPERTY_HINT_NONE, "Enable the shader cache, which stores compute shader binaries for faster loading.", true, false);
```
Zylann, line 64: Since Godot's shader cache is enabled by default, why shouldn't this one be enabled by default?

util/godot/classes/rendering_device.cpp
```diff
@@ -52,6 +52,44 @@ Ref<RDShaderSPIRV> shader_compile_spirv_from_source(RenderingDevice &rd, RDShade
 #endif
 }
 
+PackedByteArray shader_compile_binary_from_spirv(RenderingDevice &rd, RDShaderSPIRV &p_spirv, String name) {
+#if defined(ZN_GODOT)
+	// This is a copy of `RenderingDevice::_shader_compile_binary_from_spirv` because it's private.
```
Zylann, line 57: Really unfortunate that this is not exposed to modules

util/godot/classes/rendering_device.cpp
```diff
@@ -52,6 +52,44 @@ Ref<RDShaderSPIRV> shader_compile_spirv_from_source(RenderingDevice &rd, RDShade
 #endif
 }
 
+PackedByteArray shader_compile_binary_from_spirv(RenderingDevice &rd, RDShaderSPIRV &p_spirv, String name) {
```
Zylann, line 55: Since there is `shader_create_from_bytecode`, and now we have bytecode from SPIRV, then should we get rid of `shader_create_from_spirv`?
It looks like `shader_create_from_spirv` is just a shortcut in the end. I dont see a reason to keep using it or expect a different behavior from it? (referring to the "falling back to SPIRV after we already tried the SPIRV then binary way).

util/godot/classes/rendering_device.cpp
chalecampb: I get the impression they intended to do more with compute shaders, so there are oddities like this, and the unused allow_shader flag

util/godot/classes/rendering_device.cpp
chalecampb: We can remove create_from_spirv and call the two replacement functions. My goal was to maintain the existing logic (as with the fallback). If you are OK with it we can remove create_from_spirv, and then use the dual function in the noncached path.

engine/voxel_engine_gd.cpp
chalecampb: I am trying not to alter existing behavior. Existing behavior has no cache, so I left it disabled here. I can enable it by default if that's OK.

engine/gpu/compute_shader.cpp
chalecampb: As mentioned later I was trying to maintain the same code path. If we replace shader_create_from_spirv (and maybe the debugging commented out stuff) then it would be cleaner.

engine/gpu/compute_shader.cpp
chalecampb: Technically get_singleton can return nullptr before construction or after the destructor 

In cases like this, I would normally expect it to crash because there is an architectural issue where compute shaders are being compiled during startup or shutdown (probably something out of order). If you fail silently and then try again you are masking that issue. 

If you think it can never happen then there shouldn't need to be an assert. But if you think it can happen and should just retry (if there is a retry) then assert return can be used.

engine/gpu/compute_shader.cpp
chalecampb: Yeah that does sound weird, I left the tab complete text and probably should have said "falling back to create_from_spirv" which is the original method. But if we are removing that method (see later comments) then falling back doesn't really work either. My goal was to preserve the original behavior on fail.

engine/gpu/compute_shader.cpp
chalecampb: if (err != OK && err != ERR_ALREADY_EXISTS)

it works like mkdir -p

engine/gpu/compute_shader.cpp
chalecampb: The name is the name of the shader, which shows up in the cache folder 
image

engine/gpu/compute_shader.cpp
chalecampb: I sent it to the null check since it will be null if there is an error, this lets the user know

engine/gpu/compute_shader.cpp
chalecampb: Done

engine/gpu/compute_shader.cpp
chalecampb: This is just for the voxel module, if you wanted to cache some other module's compute shader binaries you would need a different implementation. As you mentioned there is no cross extension usage

For this reason I considered adding a field to add a folder (eg, voxel_shaders/zylann.voxel.compute.shaderthing) for namespacing purposes

If there's no risk of linker collision I can remove the voxel bit

engine/voxel_engine_gd.cpp
chalecampb: I was mistaken - I intended to disable this but actually just disabled the requires restart due to misread. 

It doesn't require restart so that is technically correct by mistake but was already enabled by default so no change required. 

void add_custom_project_setting(
		Variant::Type type,
		const char *name,
		PropertyHint hint,
		const char *hint_string,
		Variant default_value,
		bool requires_restart <-- false
);

engine/gpu/compute_shader.cpp
chalecampb: Made obsolete due to unifying cached and noncached load from glsl and removing the create from spirv duplication

engine/gpu/compute_shader.cpp
chalecampb: Well in all of these cases, error is just empty PackedByte... since then it can regenerate the binary and cache it.

engine/gpu/compute_shader.cpp
chalecampb: Good call

engine/gpu/compute_shader.cpp
chalecampb: Error in saving? For example if we call save_compute_shader_binary_to_cache and it opens the file for write but doesn't actually write anything or crashes right as the file was opened or something...

engine/gpu/compute_shader.cpp
chalecampb: Replaced with a boolean parameter after removing shader_create_from_spirv and commonizing that path

engine/gpu/compute_shader.cpp
chalecampb: I am sorry, you are correct. This was from a time where I combined the path and the name in that function before moving it out. Name is redundant because the path includes the folder which has the name in it.

engine/gpu/compute_shader.cpp
chalecampb: Apparently you cannot as it returns a non-writable object

engine/gpu/compute_shader.cpp
chalecampb: Chained path_join call

engine/gpu/compute_shader.cpp
chalecampb: This confused me a little as well

Engine::get_shader_cache_path sounds like it should be engine, but in practice is set by EditorPaths. I found it while searching for "standard shader processes I should be using" 

Hence, Engine::get_shader_cache_path routes to .godot in editor, and during runtime, will route to user://whatever

There is a bit of potential misalignment as there is technically no global shader path. In ShaderRD it is set by verbatim appending shader_cache and we do the same thing.

chalecampb: Edit: also I noticed despite the name, the Engine path really just points to .godot, and internally they append shader_cache to both. So I updated that for alignment purposes

engine/gpu/compute_shader.cpp
chalecampb: See other comment for shader path details

I compiled for gdextension but had not had chance to run it, 

In response to this I gave it a shot but it does not cache. It works fine just takes the normal amount of time. I will look into why.

engine/gpu/compute_shader.cpp
Zylann: There is no risk of linker collision AFAIK because it's in an anonymous namespace.

engine/gpu/compute_shader.cpp
Zylann: I mean, we don't normally expect this to happen, it would be nice to know if it does, even if it's fairly recoverable at the cost of slower load

engine/gpu/compute_shader.cpp
Zylann: > If you fail silently and then try again you are masking that issue.

It's not about failing silently, it's about logging an error and continuing (all assert/err macros log an error), similar as when the GLSL has syntax errors. If for some reason that were to happen on editor startup for example, crashing could make users unable to open their projects.

util/godot/classes/rendering_device.cpp
Zylann: Alternatively we could use `call` but that's ugly

Zylann:
I've been wondering, what clears the cache, if the user doesn't explicitely go delete the folder? Is it possible that shaders get cached more and more as their hash changes? (I'm wondering about Godot's own cache too tbh)

engine/gpu/compute_shader.cpp
chalecampb: My bad. I compiled release and it was using dev. It works fine in gdextension, the function is identical to the non-editor path (gdextension in editor routes to user/shader_cache)

chalecampb:
> I've been wondering, what clears the cache, if the user doesn't explicitely go delete the folder? Is it possible that shaders get cached more and more as their hash changes? (I'm wondering about Godot's own cache too tbh)

I have seen options menus with cache clear buttons. The godot engine similarly does not clear the shader cache. 


Modern Warfare instructions for the same 

image

Each shader cache file is 4kb to 30kb (the big one with the whole fast noise library) and only changes on device swap, driver update, or game code update (afaik). I wouldn't worry about it filling up too much.

engine/gpu/compute_shader.cpp
chalecampb: Replaced with ZN_ASSERT_RETURN_MSG

engine/gpu/compute_shader.cpp
chalecampb: I replaced all of the if/print with the ZN_ASSERT_RETURN_V_MSG

chalecampb:
I ran some tests

| Renderer | Master Tests | `compute_shader_cache` Tests | Cache Works |
|---|---|---|---|
| `vulkan` | Pass | Pass | Yes |
| `opengl3` | Fail: `test_normalmap_render_gpu` timeout | Fail: `test_normalmap_render_gpu` timeout | Yes |
| `d3d12` | Fail: `test_normalmap_render_gpu` image diff assertion | Fail: `test_normalmap_render_gpu` image diff assertion | Yes |

It works in all 3 renderers (on 4.7). The tests fail on d3d12 and opengl3 at the same spot in master and compute_shader_cache. 

The merge checks seem to have failed relating to setup concerns.

PR #872, opened 2026-06-15
Fix GPU unsupported warning timing

chalecampb:
Fixing a minor nuisance, 

**Problem Statement**
When starting the editor containing a VoxelLodTerrain using GPU generation, it gives a spurious warning

`use_gpu_generation` is enabled, but the selected renderer does not support the RenderingDevice API

And then proceeds to use GPU to run stuff anyway, since this is a timing issue. 

**Root Cause**
The function VoxelLodTerrain::get_configuration_warnings calls VoxelEngine::get_singleton().has_rendering_device()

This returns true if the GPUTaskRunner has created a local rendering device. 

In order to determine if GPU compute is supported, we need to know if the RenderingDevice API is supported. 

> Compute shaders can only be used from RenderingDevice-based renderers (the Forward+ or Mobile renderer). (from compute documentation)

So ideally this would query the earliest available rendering device - which is the global one under RenderingServer (singleton = this) if you are going to use the existence of a RenderingDevice as the basis. 

A quick timing check on master confirms this happens way before the config check

```
t_us=126070  RenderingDevice singleton assigned
t_us=3127221 RenderingServer::get_rendering_device() during VoxelLodTerrain warnings: ready
t_us=3127264 VoxelLodTerrain registers RenderingDevice API configuration warning
t_us=3240435 GPUTaskRunner first got local RenderingDevice
```

**Fix Method**
In VoxelEngine, Replaced the has_rendering_device with supports_gpu_generation().

supports_gpu_generation now just gets the RenderingServer singleton, which should by then, have a RenderingDevice singleton. This is the global rendering device and is the earliest indication (absent a feature or renderer tag or something) that GPU compute is supported. 

Since has_rendering_device and related calls were redundant, removed them. 

**Alternatives**
- I was curious if there was a more direct method in the godot engine, and there does not seem to be a flag or anything that would replace the implementation in VoxelEngine

- I considered enumerating renderers or something (finding Forward+ or Mobile enums) but this is fragile. It's cited as "RenderingDevice supports GPU compute" and separately, forward+ and mobile use RenderingDevice, but there's no guarantee this will always be the case so RenderingDevice query seems the most direct (and close to what we already have)

- VoxelTerrain doesn't have the same issue because it doesn't add the same warning. So we could also just remove the warning if it's not uniform. It may be a good time to bring it to VoxelTerrain, except, it's not clear if that one is actively maintained. 
 
**Testing Performed**

Timing analysis performed (see above), tested on 4.7 module version and 4.6.1 reference build w/ gdextension. 

bin\godot.windows.editor.x86_64.console.exe --path modules\voxel\project --rendering-method forward_plus --verbose --editor res://check_gpu_warning.tscn

supports_gpu_generation return

| Rendering Method | Module | GDExtension |
|---|---|---|
| `forward_plus` | true | true |
| `mobile` | true | true |
| `gl_compatibility` | false | false |

**Use of AI**
I used AI to search and generate timing prints, but wrote the code and PR summary myself (there wasn't much to code tbh)

**Use of Other Copyright Materials**
I didn't use any other materials at all.

### DIFF
--- a/util/godot/classes/rendering_server.h
+++ b/util/godot/classes/rendering_server.h
@@ -44,7 +44,7 @@ inline void free_rendering_server_rid(RenderingServer &rs, const RID &rid) {
 }
 
 #ifdef VOXEL_ENABLE_GPU
-inline bool supports_gpu_generation() {
+inline bool supports_rendering_device() {
 	RenderingServer *rs = RenderingServer::get_singleton();
 	if (rs == nullptr) {
 		return false;
--- a/util/godot/classes/rendering_server.h
+++ b/util/godot/classes/rendering_server.h
@@ -43,6 +43,17 @@ inline void free_rendering_server_rid(RenderingServer &rs, const RID &rid) {
 #endif
 }
 
+#ifdef VOXEL_ENABLE_GPU
+inline bool supports_gpu_generation() {
+	RenderingServer *rs = RenderingServer::get_singleton();
+	if (rs == nullptr) {
+		return false;
+	}
+
+	return rs->get_rendering_device() != nullptr;
+}
+#endif
+
 struct ShaderParameterInfo {
 	String name;
 	Variant::Type type;
--- a/engine/gpu/gpu_task_runner.cpp
+++ b/engine/gpu/gpu_task_runner.cpp
@@ -68,10 +68,6 @@ unsigned int GPUTaskRunner::get_pending_task_count() const {
 // 	return _rendering_device;
 // }
 
-bool GPUTaskRunner::has_rendering_device() const {
-	return _has_rendering_device;
-}
-
 void GPUTaskRunner::thread_func() {
 	ZN_PROFILE_SET_THREAD_NAME("Voxel GPU tasks");
 	ZN_DSTACK();
@@ -89,8 +85,6 @@ void GPUTaskRunner::thread_func() {
 		ZN_PRINT_VERBOSE("Could not create local RenderingDevice, GPU functionality won't be supported.");
 		return;
 	}
-
-	_has_rendering_device = true;
 
 	_storage_buffer_pool.set_rendering_device(_rendering_device);
 	_base_resources.load(*_rendering_device);
--- a/engine/gpu/gpu_task_runner.h
+++ b/engine/gpu/gpu_task_runner.h
@@ -79,7 +79,6 @@ public:
 	void stop();
 	void push(IGPUTask *task);
 	unsigned int get_pending_task_count() const;
-	bool has_rendering_device() const;
 	bool is_running() const;
 
 private:
@@ -87,7 +86,6 @@ private:
 
 	RenderingDevice *_rendering_device = nullptr;
 	// mutable Mutex _rendering_device_ptr_mutex;
-	bool _has_rendering_device = false;
 
 	GPUStorageBufferPool _storage_buffer_pool;
 	BaseGPUResources _base_resources;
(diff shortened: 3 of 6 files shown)


chalecampb:
Sorry didn't notice the other got merged, so I rebased and pushed so it was on latest before reviews happen.

Zylann:
This should not be called `gpu_generation`, "generation" refers to generating voxels on the GPU, but we use GPU for more than that. That's notably why it was formerly `has_rendering_device`. 

I was thinking maybe that since the check directly verifies if VoxelEngine's own device is available, the warning could get updated afterwards once the "state" actually changes. I lost count of how many times configuration warnings don't update. I often wondered about just polling configuration warnings in the editor, without having to remember to update it manually every time any cause would change. But just checking if the renderer can have an RD might be enough.

Also since the check is now "does Godot have any RenderingDevice at all", maybe it should be moved to Godot RenderingServer utilities, instead of `VoxelEngine`.

chalecampb:
I called it supports_gpu_generation because it is solely used by the warning that says "use_gpu_generation is enabled..." While "gpu generation is allowed if rendering device is supported" is what the docs say, that's one step removed from the actual question the function asks, which is "is gpu shader supported." That's the thought I put into it, at least.

But the actual name is less important to me, if you would like I can revert the name and keep the behavior. 

I checked and there is no alternative in the rendering server, or a planned feature or anything. I think it's really, technically already one call on the rendering server, it's only two to null check rendering server which might happen in headless mode or  unit testing or something. So moving the logic into RenderingServer properly would still be the one call on that object. 

I don't recommend polling or timed updates, then you get weird stuff like spurious warnings flashing and things like that. So in this case if we only fixed it by updating it after GPU runner got an RD, then we would load, see a flash of warning which was immediately cleared.

Zylann:
> I checked and there is no alternative in the rendering server, or a planned feature or anything

I'm not sure what you refer to, but if you meant "should be moved to Godot RenderingServer utilities", it's [these utilites](https://github.com/Zylann/godot_voxel/blob/master/util/godot/classes/rendering_server.h), (not having to do a PR to Godot), so that the check can remain one function call.

> I don't recommend polling or timed updates, then you get weird stuff like spurious warnings flashing and things like that

When you start having myriads of reasons you would have to update configuration warnings, it gets really tempting to avoid spaghetti just to route everything back to the node. Seeing the icon for a brief second only on startup when GPU is enabled doesnt seem like a big issue. But that was only a thought, not necessary for this PR I think.

chalecampb:
Ahhh sorry I thought you meant a PR to godot. I will update the location from VoxelEngine to the utilities.

chalecampb:
I moved it to an inline in the requested file. Still solves the issue in module and gdextension

util/godot/classes/rendering_server.h
```diff
@@ -43,6 +43,17 @@ inline void free_rendering_server_rid(RenderingServer &rs, const RID &rid) {
 #endif
 }
 
+#ifdef VOXEL_ENABLE_GPU
+inline bool supports_gpu_generation() {
```
Zylann: Can this be named `supports_rendering_device`?

generators/voxel_generator.cpp
```diff
@@ -6,6 +6,10 @@
 #include "../util/profiling.h"
 #include "generate_block_task.h"
 
+#ifdef ZN_GODOT
```
Zylann: I already fixed this

util/godot/classes/rendering_server.h
chalecampb: Absolutely

generators/voxel_generator.cpp
chalecampb: When you mentioned that, I reset back to the prior commit but was not fast enough :)

In either case I rebased to your fix and confirmed the build with both flags

PR #874, opened 2026-07-02
Fixing up a small typo in getting_the_module.md

mattmassicotte:
First, thanks so much for the work you do - it's awesome!

It's really small typo, but I noticed it while reading and thought it would be worth fixing.


Zylann:
Thanks!

PR #878, opened 2026-07-19
Add @tool and make a note of why it could be necessary

mattmassicotte:
I ran into an issue while experimenting with a script generator because I was trying to use it along with the in-editor preview. I suspect this could be a common thing when first getting started, so I think I could be worth just including as a default. Are you into that?

I also slightly expanded the smooth example, just so it is very clear what a full script would need to look like.


Zylann:
Thanks.

PR #879, opened 2026-07-19
Fix up typo and wording in instancing docs

mattmassicotte:
Tiny typo.

Also, I slightly changed the wording in the first paragraph. It does not quite read right to me, as is. However, I don't want to change the meaning or intention here. So please do let me know if this doesn't match what you were going for.

### DIFF
--- a/doc/source/instancing.md
+++ b/doc/source/instancing.md
@@ -1,14 +1,14 @@
 Instancing
 =============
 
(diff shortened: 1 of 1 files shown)


Zylann:
I'll merge both, though I might do that mistake again in the future^^" I'm not native speaker and that wording didn't occur to me as being weird or something.

mattmassicotte:
It was an extremely minor thing! Your English is *exceptionally* good. I am a life-long French learner, but I don't know if I'll ever get to the level you have.

This one requires a subject, like "it allows you..." or "it allows the system...".

Issue #597, opened 2024-02-04
`VoxelStreamSQLite` is missing chunks when loads


matthewhilton:
**Describe the bug**
When loading a stream from VoxelStreamSQLite, there are missing chunks in the save:

Possibly related to https://github.com/Zylann/godot_voxel/issues/596

**To Reproduce**
Use this example project: [VoxelSaveLoadTestSQLite.zip](https://github.com/Zylann/godot_voxel/files/14155389/VoxelSaveLoadTestSQLite.zip)
1. Change the VoxelStreamSQLite directory to somewhere on your computer
2. Start the game, and press 'E' this will generate a big cube
3. Press 'R' and check the console output, it will save the SQLite db
4. Reload the game
5. The SQLite DB is loaded, but there are missing chunks

**Expected behavior**
There shouldn't be missing chunks in the loaded edits.

**Screenshots**
The saved edit:
![image](https://github.com/Zylann/godot_voxel/assets/17095477/413d893f-369b-4062-b970-92e15cddd2ca)

The loaded edit from the SQLite DB:
![image](https://github.com/Zylann/godot_voxel/assets/17095477/77f60e94-3a94-41bc-ad75-1792df55ed82)

**Environment**
- OS: Windows 10
- Graphics card: Vulkan API 1.3.242 - Forward+ - Using Vulkan Device #0: NVIDIA - NVIDIA GeForce GTX 1060 3GB
- Godot version & Module version: 4.1.1 - am on the release https://github.com/Zylann/godot_voxel/releases/tag/v1.1.0
- Renderer used: Forward+

Zylann:
I changed the path to be just `save3.sqlite` so it saves next to the game.
I did a first test and the cube loaded properly.
I tested several more times, still could not reproduce.

Note: in this project, you have `run_stream_in_editor` turned on. That means if the scene is open, the save file will be held open and can't be deleted until the scene is closed. There can also be writing conflicts if both the editor and the game try to write into it. That didn't seem to prevent the game from loading or saving, but in order to reproduce I had to close the scene file so I could delete the save.

PR #881, opened 2026-07-21
SQLite: fix connections getting stuck inside a transaction under concurrent access

rogachevne-collab:
`Connection::open()` never sets a busy timeout, so when several connections of the same stream are open on one file at once (the pool hands them out to a background streaming thread and to `flush()` called from game code alike), a statement that can't take a lock fails immediately with `SQLITE_BUSY` instead of waiting for it.

That is worst on COMMIT: SQLite does **not** roll back a failed COMMIT, so the connection stays inside the transaction. It was then recycled into the pool as if healthy, and every later `begin_transaction()` on it failed with `cannot start a transaction within a transaction` — permanently, since nothing reset it. In practice this shows up as a sustained flood of that error under concurrent load, e.g. a first-launch generation burst on a fresh save racing an explicit `flush()`.

This addresses the two `TODO We should handle busy return codes` at the `begin_transaction()` call sites, and part of `TODO Needs better error rollback handling`:

- `sqlite3_busy_timeout()` in `Connection::open()`, so contending connections wait instead of failing outright. Kept modest (1s) because `flush()` is reachable from the main thread and exceeding it is no longer fatal.
- `Connection::rollback_transaction()`, used to bring a connection back to a usable state after a failed transaction. It also clears leftover statement state, since `sqlite3_reset()` returns the error of the *previous* evaluation, and the recovered connection would otherwise fail its next `BEGIN` once more.
- Connections that still can't be recovered are dropped instead of returned to the pool.
- Added `test_voxel_stream_sqlite_transaction_recovery`, which reproduces the stuck state deterministically (no threading involved). It legitimately prints the "cannot start a transaction within a transaction" error while passing — that's the condition under test.

**Not fixed here:** `VoxelStreamCache::flush()` clears its dirty blocks as it hands them to the save callback, before the caller's COMMIT is confirmed, so blocks caught in a failed flush are lost with no way to retry. This change makes that far rarer but doesn't close it — it seemed like it needs a design decision (unbounded cache growth if commits keep failing) rather than a quick patch, so I left it out. It may be worth a look as a candidate cause for #597, which reports missing chunks after a *normal* shutdown.

**Testing:** Windows, editor target, `precision=double`, GDExtension, godot-cpp master. The SQLite test suite passes, including the new test. Note `test_voxel_buffer_metadata_gd` fails in this configuration both with and without this change (verified against unmodified master) — it appears to be a pre-existing, unrelated issue specific to double precision. Reproduced the original error flood on a fresh save under concurrent generation + an early dig, confirmed it's gone after this change. Not tested on Linux/macOS/mono or as an engine module build.

### DIFF
--- a/streams/sqlite/voxel_stream_sqlite.cpp
+++ b/streams/sqlite/voxel_stream_sqlite.cpp
@@ -36,6 +36,18 @@ VoxelStreamSQLite::CoordinateFormat to_exposed_coordinate_format(BlockLocation::
 	return static_cast<VoxelStreamSQLite::CoordinateFormat>(format);
 }
 
+// Brings a connection back to a usable state after a transaction failed on it, and tells whether it may be reused.
+// A failed COMMIT notably leaves the transaction open, and SQLite has no nested transactions, so without this every
+// later `begin_transaction` on that connection would fail with "cannot start a transaction within a transaction".
+bool recover_after_failed_transaction(sqlite::Connection &con) {
+	if (con.rollback_transaction()) {
+		ZN_PRINT_VERBOSE("VoxelStreamSQLite: recovered connection after a failed transaction");
+		return true;
+	}
+	ZN_PRINT_ERROR("VoxelStreamSQLite: could not recover connection after a failed transaction, dropping it");
+	return false;
+}
+
 bool validate_range(Vector3i pos, unsigned int lod_index, const Box3i coordinate_range, unsigned int lod_count) {
 	if (!coordinate_range.contains(pos)) {
 		ZN_PRINT_ERROR(format("Block position {} is outside of supported range {}", pos, coordinate_range));
@@ -79,6 +91,8 @@ void VoxelStreamSQLite::set_database_path(String path) {
 		// Since Godot helpfully sets the property for every character typed in the inspector.
 		// So there can be lots of errors in the editor if you type it.
 		if (con.open(_globalized_connection_path.data(), to_internal_coordinate_format(_preferred_coordinate_format))) {
+			// Failure deliberately not handled: the connection is local and destroyed right after, so there is
+			// nothing to recover, and the path is changing anyway.
 			flush_cache_to_connection(&con);
 		}
 	}
@@ -142,7 +156,7 @@ void VoxelStreamSQLite::load_voxel_blocks(Span<VoxelStream::VoxelQueryData> p_bl
 
 	sqlite::Connection *con = con_res.connection;
 
-	const ScopeRecycle con_scope(this, con);
+	ScopeRecycle con_scope(this, con);
 
 	// Check the cache first
 	StdVector<unsigned int> blocks_to_load;
@@ -168,8 +182,10 @@ void VoxelStreamSQLite::load_voxel_blocks(Span<VoxelStream::VoxelQueryData> p_bl
 		return;
 	}
 
-	// TODO We should handle busy return codes
-	ERR_FAIL_COND(con->begin_transaction() == false);
+	if (con->begin_transaction() == false) {
+		con_scope.broken = !recover_after_failed_transaction(*con);
+		return;
+	}
 
 	for (unsigned int i = 0; i < blocks_to_load.size(); ++i) {
 		const unsigned int ri = blocks_to_load[i];
@@ -191,7 +207,9 @@ void VoxelStreamSQLite::load_voxel_blocks(Span<VoxelStream::VoxelQueryData> p_bl
 		q.result = res;
 	}
 
-	ERR_FAIL_COND(con->end_transaction() == false);
+	if (con->end_transaction() == false) {
+		con_scope.broken = !recover_after_failed_transaction(*con);
+	}
 }
 
 void VoxelStreamSQLite::save_voxel_blocks(Span<VoxelStream::VoxelQueryData> p_blocks) {
@@ -277,10 +295,12 @@ void VoxelStreamSQLite::load_instance_blocks(Span<VoxelStream::InstancesQueryDat
 	}
 
 	sqlite::Connection *con = con_res.connection;
-	const ScopeRecycle con_scope(this, con);
+	ScopeRecycle con_scope(this, con);
 
-	// TODO We should handle busy return codes
-	ERR_FAIL_COND(con->begin_transaction() == false);
+	if (con->begin_transaction() == false) {
+		con_scope.broken = !recover_after_failed_transaction(*con);
+		return;
+	}
 
 	for (unsigned int i = 0; i < blocks_to_load.size(); ++i) {
 		const unsigned int ri = blocks_to_load[i];
@@ -313,7 +333,9 @@ void VoxelStreamSQLite::load_instance_blocks(Span<VoxelStream::InstancesQueryDat
 		q.result = res;
 	}
 
-	ERR_FAIL_COND(con->end_transaction() == false);
+	if (con->end_transaction() == false) {
+		con_scope.broken = !recover_after_failed_transaction(*con);
+	}
 }
 
 void VoxelStreamSQLite::save_instance_blocks(Span<VoxelStream::InstancesQueryData> p_blocks) {
@@ -450,8 +472,10 @@ void VoxelStreamSQLite::flush_cache() {
 	}
 	sqlite::Connection *con = con_res.connection;
 
-	const ScopeRecycle con_scope(this, con);
-	flush_cache_to_connection(con);
+	ScopeRecycle con_scope(this, con);
+	if (!flush_cache_to_connection(con)) {
+		con_scope.broken = !recover_after_failed_transaction(*con);
+	}
 }
 
 void VoxelStreamSQLite::flush() {
@@ -459,12 +483,16 @@ void VoxelStreamSQLite::flush() {
 }
 
 // This function does not lock any mutex for internal use.
-void VoxelStreamSQLite::flush_cache_to_connection(sqlite::Connection *p_connection) {
+// Returns false if the transaction failed, in which case the connection may need to be recovered before reuse (see
+// `recover_after_failed_transaction`).
+bool VoxelStreamSQLite::flush_cache_to_connection(sqlite::Connection *p_connection) {
 	ZN_PROFILE_SCOPE();
 	ZN_PRINT_VERBOSE(format("VoxelStreamSQLite: Flushing cache ({} elements)", _cache.get_indicative_block_count()));
 
-	ERR_FAIL_COND(p_connection == nullptr);
-	ERR_FAIL_COND(p_connection->begin_transaction() == false);
+	ERR_FAIL_COND_V(p_connection == nullptr, false);
+	if (p_connection->begin_transaction() == false) {
+		return false;
+	}
 
 #ifdef VOXEL_ENABLE_INSTANCER
 	StdVector<uint8_t> &temp_data = get_tls_temp_block_data();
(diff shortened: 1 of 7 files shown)


Zylann:
I don't quite understand all of it, but I'd like to point out that this stream wasn't really meant to be used with multiple connections to the same DB. If I recall correctly (it's code I wrote a long time ago) it has a mutex internally to force access to only one stream function at a time. Also, streaming tasks done by the terrain are further configured to run serially (in different threads, but not overlapping, to avoid contention in the first place). So I'm not sure how you are getting multiple connections at once (several openings, at that).
There is a pool of connections mainly for two reasons:
- Well, pooling. Queries shouldn't be re-prepared every time a chunk needs to be saved
- Because historically I wanted to experiment with parallel saving, but in the end never really worked out nor got the results I wanted.

If `flush` fails and we give up, that's data loss, indeed. Not sure how that would be supposed to be handled. But again streams in use by the terrain aren't supposed to be messed with except in controlled situations where it can be proven that terrain tasks are done doing stuff with it (honestly the threading part makes it quite complicated to handle properly)

rogachevne-collab:
The mutex only guards the pool container (checkout and return). Connections are used with no lock held, and when the pool is empty, `get_connection()` opens a new one instead of waiting:

https://github.com/Zylann/godot_voxel/blob/f1141847d33f2acf5f8a70fcddbba30e5ed5bde8/streams/sqlite/voxel_stream_sqlite.cpp#L528-L553

So two connections can be live at once. In our case the game called `flush()` on the main thread while the terrain's streaming threads were still loading and saving on the same stream (fresh save, `save_generator_output` on), so the pool ran empty and a second connection was opened.

Once that happens, a COMMIT can hit SQLITE_BUSY because no busy handler is installed. A COMMIT that fails with BUSY stays inside the transaction, the connection goes back into the pool in that state, and every later BEGIN on it fails with "cannot start a transaction within a transaction". That matches the permanent error flood we hit.

The recovery part is useful even with a single connection: a COMMIT that fails for any other reason (I/O error, disk full) poisons the pooled connection the same way today. The new test reproduces the stuck state without threads.

If you'd prefer to enforce one connection at a time instead, happy to rework it that way. The rollback recovery would still be needed either way.

rogachevne-collab:
You're right about the task side, I checked: I/O tasks go through `push_async_io_task`, which enqueues them as serial, so at most one terrain task touches the stream at a time:

https://github.com/Zylann/godot_voxel/blob/f1141847d33f2acf5f8a70fcddbba30e5ed5bde8/engine/voxel_engine.cpp#L280-L283

The second user is game code. `flush()` runs on the caller's thread, outside that serial queue. And calling it is the documented way to get durability: the `VoxelStream.flush` doc says "May be used if you require all data to be written now" and points to `save_modified_blocks`. But after the save tracker completes, load tasks keep running while the player moves, and with `save_generator_output` new save tasks keep getting queued untracked. So the documented recipe (save, wait for the tracker, flush) inherently runs while the I/O thread is busy. When both touch the stream at once, the pool runs empty and a second connection gets opened. That is the situation this PR hardens.

On data loss: agreed, and this PR deliberately doesn't try to solve it, it only keeps the stream usable after a failed COMMIT. A possible follow-up would be to keep cached blocks until COMMIT succeeds, so the next flush retries them, at the cost of cache growth while failures persist.

If you'd rather make `flush()` serialize with the I/O tasks (true single connection), I can rework it that way. The rollback recovery would still be needed for plain I/O failures either way.

Zylann:
> So two connections can be live at once. In our case the game called flush() on the main thread while the terrain's streaming threads were still loading and saving on the same stream (fresh save, save_generator_output on), so the pool ran empty and a second connection was opened.

Yeah I think this was not supposed to be used like that. Though I guess it wasn't guarded for. If your change makes it work I'm ok to have it that way.

> But after the save tracker completes, load tasks keep running while the player moves

And that is why. If the goal was to save regularly, this should be done using an async function (like `save_modified_blocks`)and not something across two threads, one of which would block the main thread. And even for quitting the recipe is to trigger an async task and wait for its completion. It's ok if flush could be made to work ok somehow, but that doesn't bar against the caveats of doing so while saves are still in flight on another thread.

> On data loss: agreed, and this PR deliberately doesn't try to solve it, it only keeps the stream usable after a failed COMMIT. A possible follow-up would be to keep cached blocks until COMMIT succeeds, so the next flush retries them, at the cost of cache growth while failures persist.

Could be ok to let it grow, after all the maximum size is a threshold rather than a hard limit, but it will then keep retrying on every occasion until content gets back below that threshold. However if that happens when quitting the game for example there will be no retry to wait for.

rogachevne-collab:
Agreed on both. An async flush would be the right shape for periodic saves; the sync call blocking the main thread is a real cost of the current recipe. Could be follow-up material.

And yes, quitting is exactly why the retry idea stayed out of this PR: at shutdown there is no later flush to wait for, so keeping blocks cached only moves where the loss happens.

Happy to adjust anything in the diff if you want changes; otherwise it's ready from my side.

streams/sqlite/voxel_stream_sqlite.cpp
```diff
@@ -168,8 +182,10 @@ void VoxelStreamSQLite::load_voxel_blocks(Span<VoxelStream::VoxelQueryData> p_bl
 		return;
 	}
 
-	// TODO We should handle busy return codes
-	ERR_FAIL_COND(con->begin_transaction() == false);
+	if (con->begin_transaction() == false) {
+		con_scope.broken = !recover_after_failed_transaction(*con);
+		return;
```
Zylann, line 209: This is loading failure, but it does not set return codes, might need to log an error too?

streams/sqlite/voxel_stream_sqlite.cpp
```diff
@@ -191,7 +207,9 @@ void VoxelStreamSQLite::load_voxel_blocks(Span<VoxelStream::VoxelQueryData> p_bl
 		q.result = res;
 	}
 
-	ERR_FAIL_COND(con->end_transaction() == false);
+	if (con->end_transaction() == false) {
```
Zylann, line 232: Since this transaction only reads data, what would it mean for it to fail at the end?
If this is a loading failure, it should also set error codes

streams/sqlite/voxel_stream_sqlite.cpp
```diff
@@ -277,10 +295,12 @@ void VoxelStreamSQLite::load_instance_blocks(Span<VoxelStream::InstancesQueryDat
 	}
 
 	sqlite::Connection *con = con_res.connection;
-	const ScopeRecycle con_scope(this, con);
+	ScopeRecycle con_scope(this, con);
 
-	// TODO We should handle busy return codes
-	ERR_FAIL_COND(con->begin_transaction() == false);
+	if (con->begin_transaction() == false) {
+		con_scope.broken = !recover_after_failed_transaction(*con);
```
Zylann, line 332: This is loading failure, but it does not set return codes, might need to log an error too?

streams/sqlite/voxel_stream_sqlite.cpp
```diff
@@ -313,7 +333,9 @@ void VoxelStreamSQLite::load_instance_blocks(Span<VoxelStream::InstancesQueryDat
 		q.result = res;
 	}
 
-	ERR_FAIL_COND(con->end_transaction() == false);
+	if (con->end_transaction() == false) {
+		con_scope.broken = !recover_after_failed_transaction(*con);
```
Zylann, line 371: If this is a loading failure, it should also set error codes

streams/sqlite/voxel_stream_sqlite.cpp
Zylann, line 435: Can it happen to this one too? (arguably much harder since that request occurs on startup and never again)

streams/sqlite/voxel_stream_sqlite.cpp
```diff
@@ -450,21 +472,27 @@ void VoxelStreamSQLite::flush_cache() {
...
+// `recover_after_failed_transaction`).
+bool VoxelStreamSQLite::flush_cache_to_connection(sqlite::Connection *p_connection) {
 	ZN_PROFILE_SCOPE();
 	ZN_PRINT_VERBOSE(format("VoxelStreamSQLite: Flushing cache ({} elements)", _cache.get_indicative_block_count()));
 
-	ERR_FAIL_COND(p_connection == nullptr);
-	ERR_FAIL_COND(p_connection->begin_transaction() == false);
+	ERR_FAIL_COND_V(p_connection == nullptr, false);
+	if (p_connection->begin_transaction() == false) {
+		return false;
```
Zylann, line 539: Would it be relevant to print an error for here or would it be spurious?
Or maybe it could have a verbose print?

streams/sqlite/voxel_stream_sqlite.cpp
Zylann: This (and equivalents) might have to return a code that should be used by the caller to report when it doesnt succeed.
Depending on context it is recoverable with a warning, or should be reported with an error (like if it fails in the destructor or when changing path you really want to know).

streams/sqlite/voxel_stream_sqlite.cpp
rogachevne-collab: Done: pending blocks now get RESULT_ERROR and an error is logged.

streams/sqlite/voxel_stream_sqlite.cpp
rogachevne-collab: Realistically this fails if the connection was already in a bad state, which is what the recovery clears. There is a fork here: the results were already copied out and are valid, so failing the blocks would discard good data, but strict semantics would set RESULT_ERROR. I kept the results valid and added an error log plus the recovery. Happy to switch to RESULT_ERROR if you prefer.

streams/sqlite/voxel_stream_sqlite.cpp
rogachevne-collab: Done, same as the voxel path.

streams/sqlite/voxel_stream_sqlite.cpp
rogachevne-collab: Same question as on the voxel path above, kept the results valid here too.

streams/sqlite/voxel_stream_sqlite.cpp
rogachevne-collab: load_all_blocks itself does not open an explicit transaction (single SELECT in autocommit), so it cannot get stuck this way. But TransactionScope (used by load_meta, save_meta and migration) ignored begin and end failures, so it could. Gave it the same recovery: it tracks whether BEGIN succeeded and rolls back when END fails.

streams/sqlite/voxel_stream_sqlite.cpp
rogachevne-collab: Made it a verbose print: at that point nothing was written or dropped, cached blocks are kept and a later flush retries them. The commit failure below is the lossy one and logs an error.

streams/sqlite/voxel_stream_sqlite.cpp
rogachevne-collab: Done: flush_cache now returns success and callers report with their own context. Error in the destructor and in set_database_path, error on an explicit flush() call, warning on the automatic cache-full flush since it retries later.

streams/sqlite/voxel_stream_sqlite.cpp
Zylann: You can use `set_result_codes`

rogachevne-collab: Done. Made it an overload taking the subset of indices, since the full-span version would also overwrite blocks already resolved from the cache. Used in both load functions.

PR #882, opened 2026-07-25
Fix GDExtension import plugins losing all import options

SneaK1ng:
Looks like the GDExtension implementation builds each option Dictionary but never appends it to the output Array.

As a result, no import options show up in the editor and the generated .import file ends up with an empty [params] section, so values from _zn_get_import_options are lost.

The module version isn't affected because it writes directly into r_options.

This fixes it by adding the missing append.


Zylann:
Thanks!

PR #886, opened 2026-08-07
Fix small typo in type descrption

mattmassicotte:
Just a tiny typo I noticed.

I'm actually not 100% I'm supposed to be editing this xml manually. But I it does seem like the source used to generate final docs markdown? If there's another way, please do let me know!


Zylann:
Yes that's the source of the documentation.

mattmassicotte:
Fantastic, and thank you!

PR #896, opened 2026-09-15
Clarify that MODE_SET can also be useful for smooth voxels

mattmassicotte:
If I am understanding the documentation for using a single texture format correctly, I think this change could help clarify that  `MODE_SET` is useful for smooth voxels as well.


Zylann:
It's subtle because it is not actually useful for smooth voxels but only for the side texture data. Put it another way, it is not recommended to use that mode on the SDF channel unless you know well what you are doing with it.

mattmassicotte:
Right yes. I ended up doing this because the full documentation discusses this, but API docs do not mention it. This felt like a conflict to me. Do you think there's any point in addressing this, possibly with different wording? If not, please go ahead and close, I'm fine with that too.

Zylann:
I tend not to dwell on these things in the doc because often it would mean having to repeat things a lot in various related APIs. So I don't know, if there was something to be said I'd maybe mention the fact using this mode on the SDF channel is not recommended unless you know what you're doing.

mattmassicotte:
ok, that sounds good to me. And in that case, I'll close this.

